PR #152, opened 2015-11-12
Improve checking of Set object

luisico:
Phantomjs 2.0.0 was mislead by the way checking of Set object was done. This is a most standard way for checking existence of an Object that works across browsers.


jrobinso:
Thanks much for the pull request.  Merged.

Issue #147, opened 2015-10-12
Show length of indels in popup dialog


bw2:
When a user clicks on an insertion, show the size of the the insertion in the popup dialog.

jrobinso:
Fixed with pull request #153

PR #153, opened 2015-11-16
Added insertion summary popup

bw2:
The popup looks like

image

(from: http://exac.broadinstitute.org/variant/2-71738895-G-GGGCAGTGGTCCGAT)


bw2:
This would fix #147

Issue #185, opened 2016-04-26
Gene symbol not visible when zoomed in


keithcallenberg:
Currently the gene symbol is only being drawn in the center of the gene. If you zoom in to a region within the gene away from the center, the gene symbol is no longer visible. In desktop IGV, the gene symbol is always visible in the gene track, no matter the zoom level. Could we similarly have this in igv.js?

jrobinso:
Yes, I can see the need for that, it is more difficult in igv.js but 
clearly we need some solution.  Its probably not possible to be sure the 
gene symbol is centered, but we should be able to insure it is visible 
at all zoom levels.

In the meantime you can click on the gene to discover its name.

jrobinso:
Fixed by @skushch (#200)

PR #200, opened 2016-06-04
Feature label visibility

skushch:
Our app's users had the same feature request as is described in #185 so I figured it might be useful. 

These changes introduce the following:

- FeatureTrack's label rendering now accounts for the visible window when determining the label position. The label is now centered within the visible portion of feature.
- FeatureTrack's labels are repainted after track dragging stops - in my testing the dragging remains smooth.
- To enable this behavior I've added `trackdrag` and `trackremoved` events.

I did have one concern - do you think it's appropriate for FeatureTrack to reference its FeatureView's Tile object?

### DIFF
--- a/doc/events.html
+++ b/doc/events.html
@@ -75,6 +75,28 @@
                         <a href="../examples/events/custom-track-click.html">Popover Track Click Handling</a>
                     </td>
                 </tr>
+                <tr>
+                    <td>trackdrag</td>
+                    <td>
+                        <p>Fired whenever the user drags one of the tracks</p>
+                    </td>
+                    <td></td>
+                </tr>
+                <tr>
+                    <td>trackremoved</td>
+                    <td>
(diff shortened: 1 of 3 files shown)


jrobinso:
Thanks for taking this on!  I like the approach of redrawing the labels after dragging, but perhaps its a little more complex that it needs to be.  I think it should be fine to just force a complete repaint negating the need for the "updateFeatureLabels" function.   Calling trackView.update should accomplish this.  Does that make sense?

skushch:
I appreciate the feedback! I was focused on saving time during the repaint by only updating the visible labels. Once I compared my approach with trackView.update() it became clear that the update() call is just as fast.

I kept the calculateFeatureCoordinates() in there because that px/px1/pw calculations were being repeated in a few places. I can remove it if you like.

jrobinso:
OK, merged.  I did push a minor tweak, to do the repaint when dragging has ended, rather than during the drag operation.

Issue #137, opened 2015-08-04
Show vertical center line


bw2:
Would it be possible to show the center line to help line up variants with reads?

jrobinso:
This is already an Alignments preference option (View > Preferences > 
Alignments).

jrobinso:
Apologies I see this is an igv.js ticket,   not IGV.   Yes we will add 
this option.

THanks

bw2:
Hi Jim, I'm now adding IGV.js to another portal and wanted to revisit this.
What do you think about an API for this being like:
`v1 = igv.browser.add_vertical_line("chr1:12345") `

and, possibly (though not as needed) a remove method like:
`igv.browser.remove_vertical_line("chr1:12345")` or  `igv.browser.remove_vertical_line(v1)`

The line would then appear across all tracks and scroll left-right with the other features.
Even better, these could be track-level functions, and could be called in a loop on each track when a global line is needed.

subji:
+1

vingorius:
+1

bw2:
Hi Jim, if I have some time to work on this, would you be open to a PR on this?

jrobinso:
Yes, for sure.   I will have time myself after Wednesday.  I've been 
traveling basically the whole month.

jrobinso:
This one might be tricky, btw,  I'm not sure we have an element  to draw 
the line on.   Perhaps you should study it and propose something first,  
or just go for it and we'll iterate,  but I have a feeling this will 
take a few attempts.

bw2:
Got it. Without looking into the details yet, I was thinking to draw it on the track(s) so the line would be fixed at a particular genomic coord and  scroll horizontally along with the feature it's marking.

jrobinso:
Comment -- pull request #207 does not really address this case directly, it is more relevant to #131.

turner:
Closed with commit de0b9e3a

turner:
See `igv-dev.html `for configuration options:

To present fixed one-base wide guide pair:
`showCenterGuide: true`

To present cursor tracking guide:
`showCursorTrackingGuide: true`

Issue #131, opened 2015-06-22
[FEATURE REQUEST] - Have vertical line that goes across tissue tracks, gwas catalog track, and gene track


khadley:
It would be helpful if a person moused over the genome browser and had a vertical line that was drawn across all of the tracks (tissue, gwas catalog, and gene) so that the user can see what lines up together.

The mouse hover to draw line functionality is similar to how it's drawn on this plot on tumor portal:

http://www.tumorportal.org/view?geneSymbol=TP53

vingorius:
+1

jrobinso:
I'll try to get to this soon.  Coincidentally I developed the tumor portal viz referenced above.   I never liked the "Straightedge cursor" label and checkbox,  any feedback or suggestions on how to enable / disable the line?

vingorius:
how about "guide ruler" or "guide line" as label?

And  it has to be distinguished with 'center line' which was mentioned at #137 .
Color or style(dotted or line) may be good.

nlwashington:
+1

jrobinso:
This is partially implemented with #207.   What's missing is a button to toggle it dynamically.

PR #207, opened 2016-06-20
Vertical guide line implementation

skushch:
The changelist introduces a centered vertical line that can be enabled using `showVerticalLine = true` browser config #137. Additionally, the line may be anchored to the cursor if `showVerticalLine = "sticky"` is set #131.

The line itself is just a border on an absolutely positioned DIV within the contentDiv. Its appearance can easily be styled using CSS and it avoids any additional canvas repainting.

### DIFF
--- a/css/igv.css
+++ b/css/igv.css
@@ -1390,6 +1390,16 @@ div.ui-dialog fieldset {
     height:100%;
 }
 
+.igv-vertical-line-div {
+    position: absolute;
+    border-left: 1px dashed #999;
+    top: 0;
+    bottom: 0;
+    right: 50%;
+    width: 1px;
+    z-index: 1;
+}
+
 .igv-content-canvas {
     position: absolute;
     width: 100%;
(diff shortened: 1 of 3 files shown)


jrobinso:
Hi, once again nice work.  It very neatly satisfies #131.   However, #137 is really about framing a specific base, visually with a pair of lines on either side.   Its more about highlighting a specific base  (usually a snp).

jrobinso:
I did push a modification, making the vertical line a child of trackContainerDiv.  The ideogram is not really aligned with the tracks,  it covers the entire chromosome always,  so having the line extend into it could be misleading.

jrobinso:
Some final modifications, 
Configuration option is now "showGuideLine".  There will eventually be other vertical lines, specifically around specified bases (#137).
Default is now "sticky" (follow mouse).  To fix it use configuration value "center".
There is a new button in the control bar to toggle the state.

PR #236, opened 2016-08-15
Allow HTML in track labels. Fix #235

### LINKED ISSUE #235: Allow html in track labels
An older version of igv.js set track labels using innerHTML, so I started adding font-awesome icons to them - for example
image

The latest code uses .text(..). Can it be switched back to allowing html?

The following changes were sufficient to do this, but a new config attribute like `nameWithHtml` might be better? 

If you agree I could submit a pull request.

```
--- a/js/trackCore.js
+++ b/js/trackCore.js
@@ -71,7 +71,7 @@ var igv = (function (igv) {

         track.name = label;

-        $(track.trackView.viewportDiv).find('.igv-track-label').text(track.name);
+        $(track.trackView.viewportDiv).find('.igv-track-label').html(track.name);

         if (track.trackView) {
             track.trackView.repaint();
diff --git a/js/trackView.js b/js/trackView.js
index d727192..74421fb 100644
--- a/js/trackView.js
+++ b/js/trackView.js
@@ -191,7 +191,7 @@ var igv = (function (igv) {
             description = this.track.description || this.track.name;
             $trackLabel = $('<div class="igv-track-label">');

-            $trackLabel.text(this.track.name);
+            $trackLabel.html(this.track.name);
```

bw2:


jrobinso:
@turner -- Any reason not to do this?

turner:
This looks fine for the restricted case of adding font awesome or similar. Note: there is an implicit assumption the label is a one-liner.

Issue #237, opened 2016-08-17
CORS error on portals.broadinstitute.org/webservices/igv/locus


david-ma:
Searching for a feature, e.g. TP53 relies on this code:

`            this.searchConfig = {`
`                // Legacy support -- deprecated`
`                type: "plain",`
`                url: "https://portals.broadinstitute.org/webservices/igv/locus?genome=" + genomeId + "&name=$FEATURE$",`
`                coords: 0,`
`                chromosomeField: "chromosome",`
`                startField: "start",`
`                endField: "end"`
`            }`

Unfortunately it is giving a CORS error right now.

jrobinso:
Thanks for the report, looks like a server-side error.  Do you know when this began happening?

david-ma:
We only noticed it yesterday, but it may have started earlier.

jrobinso:
Actually I don't get the error,  you'll have to provide more details.  Try our examples at igv.org,  or just enter this in a browser with developer tools on.   You'll see the CORS headers in the response.    https://portals.broadinstitute.org/webservices/igv/locus?genome=hg19&name=egfr
Can you give me something, e.g. a web page, that reproduces the error you are seeing?

jrobinso:
Capturing the response with developer tools from your app might be enough.

david-ma:
Ok I figured it out, I had an old version of igv.js (built on the 25th of July, 2016) which was using this:

url: "//www.broadinstitute.org/webservices/igv/locus?genome=" + genomeId + "&name=$FEATURE$"

Turns out it wasn't happy with the redirect. Updating the url fixes this problem.

david-ma:
I've also noticed that I can pass in my own search url in the igvOptions

search: { url: "/PathOS/Transcript/locusJSON?name=$FEATURE$" }

Is this documented anywhere?

It seems to require a JSON that looks something like this:

`[{`
`"chromosome":"chr17",`
`"start":"7572927",`
`"end":"7579912",`
`"type":"refseq"`
`}]`

However, this doesn't work because in "handleSearchResult", when it is applying flanking, it treats "igv.browser.flanking" as a string and then concatenates the string it to the result, rather adding it.

I'll fix this and submit a pull request.

jrobinso:
[...]

PR #238, opened 2016-08-17
#237 parseInt on "start" and "end" in handleSearchResult

david-ma:
When using a JSON to pass back search information, if the "start" and "end" are passed back as strings instead of integers, they must be parsed or there will be errors.

### DIFF
--- a/js/browser.js
+++ b/js/browser.js
@@ -810,11 +810,11 @@ var igv = (function (igv) {
         igv.browser.selection = new igv.GtexSelection('gtex' === type || 'snp' === type ? {snp: name} : {gene: name});
 
         if (end === undefined) {
-            end = start + 1;
+            end = parseInt(start) + 1;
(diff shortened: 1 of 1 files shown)


david-ma:
Note that if the JSON is passed back with "start" and "end" as integers instead of strings, it'll be fine without this fix.

This is not obvious from the current documentation.

david-ma:
I guess this is fine and doesn't need to be fixed.
I have made changes to the wiki that will hopefully prevent other users making this mistake: https://github.com/igvteam/igv.js/wiki/Browser

jrobinso:
Thanks, I had also added a note about types to the wiki.   I had not closed
it because I intended to test to see if Integer.parseInt() would blow up or
do the wrong thing if passed an actual integer.    But its probably better
to specify the types in the doc than try to be too clever.

PR #267, opened 2016-09-12
self reference to parser added

zoecastillo:
FYI: This solved the issue of the wig/bedGraph files not being displayed for us.


jrobinso:
That really shouldn't be necessary, parser is defined in the enclosing function scope.   Do you have a bedgraph or wig file to reproduce the original issue?  I don't mind merging the pull request, its redundant but not harmful,  however I want to understand the original issue.   If I've missed something let me know.

jrobinso:
Never mind,  I see the issue, very subtle.

jrobinso:
Can I ask what browser & browser version you are using?

PR #273, opened 2016-09-29
bug fix about choosing first search result when multiple returned

JiaoJiao123:
The original design is to use the first one when multiple search results found. But there were something wrong with the logic implemented in the code. 

@jrobinso Is there any update about the function presentSearchResults to allow user to choose?

Another thing is, I customized some parts of jgv.js for cBioPortal usage. There are few features added. We are thinking to send pull request back to igv team. May I know should I send pull request back to master or other branch?


jrobinso:
Thanks for the fix.  Just released in version 1.0.5

PR #274, opened 2016-09-29
Support GCS paths for all track types. Fixes #271.

### LINKED ISSUE #271: Cannot specify GCS path for BAM indexURL
gs:// paths do not work correctly when set in the indexURL field of a bam track. This appears to be due to [this line](https://github.com/igvteam/igv.js/blob/master/js/bam/bamReader.js#L42). The logic should probably be:
```
this.baiPath = config.indexURL || config.url + ".bai";
this.baiPath = 'gcs' === config.sourceType ? igv.translateGoogleCloudURL(this.baiPath) : this.baiPath;
```

There appears to be a similar issue with the headURL.

mikeas1:
Let me know what you think. I believe this should be sufficient to allow any track type (references, variants, etc) to support gs:// paths.

### DIFF
--- a/js/bam/bamReader.js
+++ b/js/bam/bamReader.js
@@ -33,13 +33,9 @@ var igv = (function (igv) {
 
         this.filter = config.filter || new igv.BamFilter();
 
-        this.bamPath = 'gcs' === config.sourceType ?
-            igv.translateGoogleCloudURL(config.url) :
-            config.url;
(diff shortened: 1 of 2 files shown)


jrobinso:
Looks good.   I'm working on the oAuth problem now.

jrobinso:
Fix included in v 1.0.5

PR #276, opened 2016-10-02
Return loaded tracks from loadTrack and loadTracksWithConfigList. Fixes #275

### LINKED ISSUE #275: FR: Return track created by loadTrack
It would be convenient to have a reference to the track created and added by loadTrack in order to be able to later remove the track (and possibly re-add the same track later without having to construct it from scratch).

The intended behavior would be that the following works:
```
// Add a new track to the browser.
var newTrack = browser.loadTrack(trackConfig);
// Remove that track from the browser.
browser.removeTrack(newTrack);
// Add the same track back to the browser. Note the usage of addTrack vs. loadTrack.
browser.addTrack(newTrack);
```

mikeas1:
Let me know what you think. I haven't fully reasoned about whether re-adding a track that was previously removed will completely work, but from my initial look it seemed possible.

### DIFF
--- a/js/browser.js
+++ b/js/browser.js
@@ -118,16 +118,17 @@ var igv = (function (igv) {
     igv.Browser.prototype.loadTracksWithConfigList = function (configList) {
 
         var self = this;
+        var loadedTracks = [];
 
         configList.forEach(function (config) {
-            self.loadTrack(config);
+            loadedTracks.push(self.loadTrack(config));
         });
 
         // Really we should just resize the new trackViews, but currently there is no way to get a handle on those
         this.trackViews.forEach(function (trackView) {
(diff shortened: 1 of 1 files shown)


mikeas1:
Merged separately in 584ff9134e760c9768816489fee9c35fe02b6456.

jrobinso:
I think that's the right thing to do (return the newly created tracks).   I
am not sure that re-adding will work,  it should but has not been tested.
If this turns out to be problematic we could add  "hide() / show()"
 methods to accomplish the same thing.

Note also there is a findTrack method in the master, not released or
documented yet.    This method will enable you to find tracks based on any
track property.   It was added to enable bulk operations on groups of
related tracks.

PR #277, opened 2016-10-03
get segment data in JSON format from API call

JiaoJiao123:
In cBioPortal, we get segment data from API calls, where the data is returned in JSON format. We added the functions below to support this feature, and in the hope that this feature could be useful for igv.js and be utilized for other similar usage.

It could be called easily by adding values in the initial options setting like below:
//json: flag to control using API call or not (true or false)
//url: the API url you want to access
//method: how you want to access your API
//parameters: parameters you want to pass in your API call 
options = {
                    ...
                    tracks: [
                        {
                           ...
                            type:"seg",
                            url: "http://www.cbioportal.org/beta/api-legacy/copynumbersegments",
                            json: true,
                            method: "POST",
                            sendData: 'cancerStudyId=coadread_tcga_pub&chromosome=12'
                        },
                        ...
                    ]
                };

### DIFF
--- a/js/feature/featureFileReader.js
+++ b/js/feature/featureFileReader.js
@@ -105,16 +105,24 @@ var igv = (function (igv) {
             };
  
             function parseData(data) {
-                self.header = self.parser.parseHeader(data);
-                if (self.header instanceof String && self.header.startsWith("##gff-version 3")) {
-                    self.format = 'gff3';
+                if(self.config.json){
+                    fulfill(data);
+                }else{
+                    self.header = self.parser.parseHeader(data);
+                    if (self.header instanceof String && self.header.startsWith("##gff-version 3")) {
+                        self.format = 'gff3';
+                    }
+                    fulfill(self.parser.parseFeatures(data));   // <= PARSING DONE HERE
                 }
-                fulfill(self.parser.parseFeatures(data));   // <= PARSING DONE HERE
             };
 
 
             if (self.localFile) {
                 igvxhr.loadStringFromFile(self.localFile, options).then(parseData).catch(reject);
+            }
+            else if(self.config.json){
+                //this is customized for cBioPortal use case
+                igvxhr.loadJson(self.config.url, self.config).then(parseData).catch(reject);
             }
             else {
                 igvxhr.loadString(self.url, options).then(parseData).catch(reject);
--- a/js/igvxhr.js
+++ b/js/igvxhr.js
@@ -154,7 +154,7 @@ var igvxhr = (function (igvxhr) {
             }
         });
     }
-
+    
     igvxhr.loadArrayBuffer = function (url, options) {
 
         if (options === undefined) options = {};
@@ -166,7 +166,13 @@ var igvxhr = (function (igvxhr) {
 
         var method = options.method || (options.sendData ? "POST" : "GET");
 
-        if (method == "POST") options.contentType = "application/json";
+        if (method === "POST"){
+            if(options.json){
+                options.contentType = "application/x-www-form-urlencoded";
+            }else{
+                options.contentType = "application/json";
+            }
+        } 
 
         return new Promise(function (fulfill, reject) {
 


JiaoJiao123:
@jrobinso could you review this pull request when you get a chance? Thank you!

jrobinso:
I won't have time for a full review today.   However, could you resubmit
without the new method in igvxhr?   All xmlhttprequests must go through
igvxhr.load.   See igvxhr.loadJson, you should be able to use that if not
let me know why.    There should be no use of $.ajax,  $.when,  or $.then
methods.

JiaoJiao123:
@jrobinso I remember I tried loadJson earlier and it didn't work, but I forgot why. I will try again and let you know.

JiaoJiao123:
@jrobinso I tested it and realize the error of accessing segment API is caused by the contentType setting. After I set it to "application/x-www-form-urlencoded" and it is fine. So I removed the new method of doing AJAX call and updated the PR.

jrobinso:
Thanks,  but looking forward to multi-locus view this API is not going to work.   With the example you have given it looks like you are returning all the data for a chromosome in a single call for a track.   If you are viewing genes on different chromosomes what are you going to do?

jrobinso:
The right solution for this is to have a webservice that takes a genomic range and returns the data for that,  this would be called multiple times for multi-locus view.   It would be o.k. if the webservice used the chromosome only, and ignore the other information.  Does that make sense?

JiaoJiao123:
@jrobinso you are exactly right. we did have problem of not able to visualize gene on another chromosome. And then I add some condition checking when user try to switch to another genomic range or gene. I check the returned chromosome value after done searching, if it is different with the previous chromosome, we will trigger the segment API call again to fetch data. I will send a PR about that part and show you what I did.  

It is implemented in the [cBioPortal beta version](http://www.cbioportal.org/beta/index.do?tab_index=tab_visualize&cancer_study_list=coadread_tcga_pub&cancer_study_id=coadread_tcga_pub&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=coadread_tcga_pub_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=coadread_tcga_pub_gistic&Z_SCORE_THRESHOLD=2.0&case_set_id=coadread_tcga_pub_nonhypermut&case_ids=&gene_list=KRAS+NRAS+BRAF&gene_set_choice=user-defined-list&Action=Submit&show_samples=false&).  Click on the link and you can find IGV in the CN Segment tab.

jrobinso:
Very nice integration job, but Its quite unlikely I will be able to accept
the pull request,  this is a general viewer for many portals not something
specific for cBio.    I have communicated with you the type of web service
interface we can accept into the igv.js project.    It should be very easy
for you to support that on the server side,  rather than have special cBio
only switches in the client code.

Are you going to stick with the tabbed solution for multiple genes?  That
is a fine solution, we will release the multi-gene view within the next
month of so in any event.

Jim

JiaoJiao123:
@jrobinso I think we are releasing this cBioPortal beta version this week. But it would be great if you could let us know when the multiple genes view is released. We will decide then. 

> I have communicated with you the type of web service interface we can accept into the igv.js project

If you have time, could we do a hang out chatting sometime? It is easier to communicate that way. I might not totally get what you said easier. It doesn't matter very much this PR get merged or not, but I think it is a very important feature for IGV library to support JSON data through API directly.

jrobinso:
We do supporrt JSON data,  that is not the issue, many websites using
igv.js have services that return data to igv.js in json format.   The issue
is your interface,  we need a service that accepts genomic coordinates as
an argument and returns the json data.    That's the basic model.    I'm
not sure why this is difficult to get across,  if you think a web hangout
would be helpful I can do that anytime  (I am on west coast time).

Jim

JiaoJiao123:
Thank you very much! I can do the talk now. You can use google hang out to call me at jiaojiaowanghere@gmail.com. or let me know yours and I can call you

jrobinso:
I just sent a google invite

JiaoJiao123:
@jrobinso for some reason, it showed up in my email not in invite. I just sent you a google hangout invite. did you see it?

PR #278, opened 2016-10-03
sort sample by segment value at first initialization

JiaoJiao123:
We would like to have the segment viewer sorted decreasingly by the segment mean value at first load.  
To use this, we need to add specify [type: "seg"] in the segment track.


jrobinso:
This would be better treated as a feature request than a pull request.  I understand the intent.   However, be aware that in multi-gene view (coming soon) that you cannot sort each individual gene in a track individually,  the samples need to be in the same order otherwise the views are not comparable.   This is in fact one of the main uses of multi-gene view,  sort based on CN in one gene and see if it is correlated in others.    So setting an initial sort does not make sense in this case.

JiaoJiao123:
ok get it. I will make a feature request about the sorting for single gene view.

Issue #303, opened 2017-01-10
Handling css and img files


david-ma:
Hi Jim and co,

I've noticed that there are a few css and img files that may or may not be needed by igv.js

The wiki recommends hotlinking the css file on igv.org, is that still the correct/recommended practice?
https://github.com/igvteam/igv.js/wiki/Dependencies

---

If igv-gtex.css, igv.css and opentip.css are required, I recommend using grunt to concat and also concat+minify these files to the dist folder on build. I also recommend copying css/img to dist/img on build.

I will create a pull request that does this.

david-ma:
It might also be nice to use SASS, but that's probably overkill if you aren't going to do a lot of CSS stuff.

jrobinso:
Hi,  igv-gtex.css and opentip.css are not listed in dependencies, and not
used for any examples.    I agree with css/img,  I actually thought that
was copied.

jrobinso:
Also, David, where do you see a recommendation to hotlink?  We are agnostic
on that,  you can use the links as provided or download the files yourself.

david-ma:
Oh sorry, it doesn't say "recommended" in those words, but it kind of implies it because of this line: https://github.com/igvteam/igv.js/wiki/Dependencies#igv-css

and also the examples (e.g. https://github.com/igvteam/igv.js/blob/master/examples/1kg-vcf.html ) refer to the igv.org hosted files.

sgratzl:
within https://github.com/datavisyn/igv_wrapper I added a simple example how to bundle everything together using webpack

jrobinso:
Thanks @sgratzl -- that's very helpful.

PR #304, opened 2017-01-10
#303 concat css, concat and minify css, and copy img to dist folder when building with grunt

david-ma:
See issue #303

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -27,6 +27,10 @@ module.exports = function (grunt) {
                     'vendor/underscore-min.js'
                 ],
                 dest: 'dist/igv.js'
+            },
+            css: {
+                src: 'css/*.css',
+                dest: 'dist/igv.css'
             }
         },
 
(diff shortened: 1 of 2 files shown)


jrobinso:
Hi,  the igv-gtex.css and opentip.css files are not used, and shouldn't be included.  Where did you see references to them?   Also, we had issues after minimizing the css with grunt in the past,  that's why we don't do it.   What is the advantage?  Its a tiny file already.   Minimizing it is o.k. as long as the unminimized version is also copied.

david-ma:
OH! Ok, this commit is unnecessary then.

1. I wasn't sure which .css files were needed, so I was including all of them just to be safe.
2. With multiple .css files, concatenating them would be a good thing to do.
3. After concatenation, adding the minification step is trivial.

---

I still think it would be good to copy the igv.css and img/* stuff to the dist folder on running grunt, so it is clear which files should be distributed.

jrobinso:
Yes, agreed.  You can submit a PR for just that, or open a git issue and we
will do it.  I'm on IGV desktop for the next few weeks.

david-ma:
Ok cool, I'll make a new PR tomorrow. This one can be closed.

Issue #324, opened 2017-02-13
bigwig - duplicate features for leafItems[0]


blajoie:
Hi,

For awhile now, we (@nhjohnson) have been noticing a visual artifact with some bigwig data, it was only happening intermittently and only when plotting data as a line...

e.g. bigwig artifact (notice the horizontal lines)
https://s3-us-west-2.amazonaws.com/ilmn.igv-test/igv-bigwig-bug.png

It turns out there was a small bug in the BWSource code, when the number of leafItems was > 1.
The first leaf was being duplicated in the resulting merged list of all features, which was causing a plotting artifact in the sorted list of all features.  This only manifested when plotting as lines because the point/rectangle dupes were plotted on top of one another.

After the fix, data now looks AOK and the duplicates are gone.
https://s3-us-west-2.amazonaws.com/ilmn.igv-test/igv-bigwig-bugfixed.png

Will submit a PR with the fix.

PR #325, opened 2017-02-13
bug fix for bigwig parsing, if leafItems/featureArrays was > 1, then …

blajoie:
See issue here.
https://github.com/igvteam/igv.js/issues/324


jrobinso:
Thanks for the PR.  Open source is great.

PR #357, opened 2017-04-26
encode-dev.html. Example using EncodeTable for igv.js.

turner:
encode-dev.html. Example using EncodeTable for igv.js.


jrobinso:
Doug, why all the "cursor" files?

PR #392, opened 2017-08-16
Authorization and heterozygosity display bugfixes

ishanley:


jrobinso:
Accepting as a legacy option,  but there is a more recent general method to add oauth tokens.

PR #400, opened 2017-09-06
arrayToCommaString bug

liub1993:
This function can be simply replaced by `join()`,  by the way, **`1` was used as `i`**.


jrobinso:
Nice, thanks,  I'm new to this language.

PR #407, opened 2017-09-12
Isaac dev

ishanley:
Apart from updating the color scheme and doing the plink file uploading and parsing, there were a couple bugfixes:
- Clicking on the variant was not displaying all info values.  
This was fixed by not calling arrayToCommaString(variant.info[key]) since variant.info[key] was already a string and not an array.

- Clicking on the call for str variants didn't display str-specific attributes.  
This was fixed by changing the check for variant type from 'this.type === "str"' to 'variant.str' in variantTrack.


jrobinso:
Is variant.info[key] always a string?  It might need a test (array or string).   I'm not certain.

jrobinso:
Why the change from  type = 'str'  to variant.str?   I think we want to use "type = 'str'",  there are many types of variants not just 2 (str and not str).

ishanley:
OK, we can use a test for variant.info[key]. The problem was that when it wasn't an array, the method arrayToCommaString was returning an empty string which is why nothing was showing. We could just make it return the input if it's not an array.
For detecting variant type, we can use variant.type instead of variant.str. Just for that popup, it was using this.type in variantTrack which wasn't even populated, so the str attributes weren't showing up.
I can go ahead and make these changes.

jrobinso:
There should be a default variant.type,  or treated undefined or "" as the default.

jrobinso:
if variant.type is undefined this will fail:  variant.type === 'str'

This is safer     'str'  === variant.type

ishanley:
Oh ok.

PR #410, opened 2017-09-18
Isaac dev

ishanley:
Can now group callSets in a vcf based on attributes loaded from igv.sampleInformation.


jrobinso:
Be careful with tests like this for numeric attributes, zero evaluates to "false".   So if someone tries to set a gap of zero it will become 10.    Not likely but still.

PR #451, opened 2017-10-26
validate/expose magic from BAM header

jmtcsngr:


jrobinso:
Perfect, thanks for the improvement.

PR #453, opened 2017-10-27
Start removing shared.js from examples and dev.

turner:
This is a maintenance thing. Support for Inclusion of shared JS is going away in 2018 for - at least - Chrome browser.


jrobinso:
Also, share does not work with phantom / qunit

PR #460, opened 2017-11-01
Sequence translate

jrobinso:


jrobinso:
This is merged to master,  thanks again.

PR #476, opened 2017-11-11
 Set headers for GA4GH GET requests

mschroering:
I was using this against a local GA4GH service and noticed that the Authorization header was not getting set on GET requests.   I only found some old GA4GH tests that may cover this change.   I was not sure if there was some other test I should verify here, but the change is pretty minor.


jrobinso:
Accepted as its consistent with our current global setting,  but the method for setting the token might change in the future.   An alternative means for providing the token is described here:  https://github.com/igvteam/igv.js/wiki/OAuth-Support--(beta-only)

PR #505, opened 2017-12-15
Wig track data range issue. Now using trackView.setDataRange when toggling autoscale.

turner:
Wig track data range issue. Now using trackView.setDataRange when toggling autoscale.


jrobinso:
Tests like this will fail with numbers
if(min)
    self.min = min

0 is a valid value for min (actually a common value),  but is "falsey",  to its impossible to set min to zero.

PR #516, opened 2018-01-11
Use standard oauth features for ga4gh

mschroering:
This is a follow up to my last PR to use the standard oauth features for ga4gh where one can provide a function that resolves to a token.

### DIFF
--- a/js/ga4gh/ga4ghHelper.js
+++ b/js/ga4gh/ga4ghHelper.js
@@ -30,9 +30,9 @@ var igv = (function (igv) {
      * @param options
      */
     igv.ga4ghGet = function (options) {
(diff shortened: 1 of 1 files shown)


jrobinso:
This doesn't appear to be a benign change,  what happens to sites that currently use the global

igv.oauth.google.access_token

?

mschroering:
Good point.  I changed the code to allow it to fallback and use `igv.oauth.google.access_token` like it used to.  Now it sends the token in the `oauthToken` option parameter to `igv.xhr.loadJson` which will evaluate if it's a function or a string.

jrobinso:
Looks good

PR #523, opened 2018-02-13
fix #522

### LINKED ISSUE #522: igv-beta.js - typing any chrX locus into the search bar results in "Unrecognized locus" error
This occurs on hg19 and hg38 

image

bw2:


jrobinso:
Thanks for the fix,  I fixed it in a different way.   I'm trying to be case insensensitive

PR #529, opened 2018-02-22
Bug fix. bamSource.js.

turner:
This bug caused loading of local indexed BAM files to not work.


jrobinso:
I don't understand the fix, or the problem for that matter.

PR #534, opened 2018-02-26
Repair Track AutoHeight.

turner:
Repair Track AutoHeight.

### DIFF
--- a/dev/igv-dev.html
+++ b/dev/igv-dev.html
@@ -344,8 +344,11 @@
         options = {
             // locus: 'all',
             // locus: '1',
-            // locus: 'SLC25A3',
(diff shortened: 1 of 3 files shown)


jimhavrilla:
is this in reference to #524?  Thanks!

jrobinso:
This should be fixed now.  Pull and give it a try.

jimhavrilla:
image

So now the height DOES autoexpand when adding autoHeight: true but the scroll bar seems to not be there anymore.  Must be a bug.  Because there are more transcripts in the Squished view and I cannot scroll.

image

jimhavrilla:
In reference to #524

PR #546, opened 2018-03-06
Str dev

ishanley:
- fixed bug where STR variants were not showing on variantTrack
- now checks for "VT" info tag in VCF to determine type of variant
- when callsets are grouped by familyId in the variantTrack, each family is now ordered (father, mother, then children)
- fix problem where height of variantTrack was sometimes wrong when callsets were grouped


jrobinso:
This is not valid.   All deletions will have referenceBases.length > 1,  and many other types of variant as well.   If its not explicitly tagged an STR then its not an STR.   This needs to be removed.

+       if (variant.referenceBases.length > 1) {
             variant.type = 'str';
         }

Have these changes been well tested with non-STR vcfs?

Issue #545, opened 2018-03-06
Retina display support


romanzenka:
Hello,

I am using IGV on a Mac and I'm noticing blurry display thanks to not using the retina display.

I would be interested in contributing support for retina.

Is this something you are at all interested in / willing to support long term?
I believe these displays will only be more prevalent over time.

In order to keep display fidelity, I would make larger canvas, scale it down using CSS, then add a transformation to all current IGV drawing code that would make the drawings 2x larger in case of a retina display.

This would ideally be concentrated in the `igv-canvas.js` file as not to create large amount of code changes all across IGV.

Over time this can be optimized by making the IGV code dpi-aware - I believe that some of your visualizations would benefit from having 2x or even 3x resolution. Smartphones are definitely a platform that would benefit from this.

In case that this is not at all interesting to you, there is an option to use `image-rendering: pixelated` as the style on the canvas element that somewhat reduces the blur.

Before: 
![image](https://user-images.githubusercontent.com/66568/37064728-c23c808a-2163-11e8-97c9-115cbee97e73.png)

After:
![image](https://user-images.githubusercontent.com/66568/37064761-db555ea2-2163-11e8-800f-1efabf3b77c9.png)

romgrk:
I 100% support @romanzenka, I've been thinking about this yesterday and I'm including here some things I found:
 - Canvas 2d text drawing will always be less perfect than DOM. It might be interesting to either render the text content through DOM, or through WebGL. WebGL could also be used to render lines, as it's made to be pixel perfect (albeit it's like shooting a target with a nuke)
 - [https://stackoverflow.com/a/40074278/6303229](https://stackoverflow.com/a/40074278/6303229) A good resource to understand how what is described by @romanzenka works

jrobinso:
I haven't had time to study this yet.   Text is fuzzy on canvas, a solution
to that would be welcome.   We did an exhaustive performance study of
Canvas vs DOM approaches (mainly SVG) at the beginning of the project and
canvas won hands down for our use cases.   I think other similar projects,
including an IGV knock-off called "pileup.js" as well as "dalliance" used
an SVG approach initially and switched to canvas later for performance.
We support many track types and use cases,  not all of which you
necessarily use,  but at this point canvas based rendering is baked in and
is not going to be changed in the foreseeable future.

All the developers on the project (both of us) have Macs with Retina
screens and I haven't noticed any fuzziness other than with text.

I will respond to @romanzenka when I've had time to look at the issue,  in
[...]

PR #553, opened 2018-03-12
Only reinitialize canvas when resizing (#545)

romanzenka:
I think it is better to do expensive operations only when you really must. My assumption is that changing canvas backing store size means the browser has to reallocate the entire canvas buffer, which really should be done only on resizes...

### DIFF
--- a/js/ideogram.js
+++ b/js/ideogram.js
@@ -279,7 +279,7 @@ var igv = (function (igv) {
 
             var shim = 1,
               ideogramTop = 0,
-              shim2 = 2;
+              shim2 = 0.5;
 
             if (!igv.browser.genome) {
                 return;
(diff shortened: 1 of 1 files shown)


romanzenka:
I made the ideogram looks like this on retina which is more on par with what the low-res version did:

image

This is still sub-optimal (overlaps of the original boxes with the outline), I will probably play with this a tiny bit to get the pixels just right.

jrobinso:
OK thanks.  Also verify that it still looks o.k.,  or not any worse,  on
low-res  (devicePixelRatio == 1).

romanzenka:
@jrobinso I have two monitors, one retina, one normal, connected to the same computer, so I always verify both.

romanzenka:
I pushed this .. not sure what your original intent was here...

image

romanzenka:
image
image

jrobinso:
Very nice

jrobinso:
I'm working on the dpi for tracks, which is all in viewport.js.   Its a little more complex, and we are doing other development there,  so it will be a few days.  But thanks for the ideogram implementation,  it provided the model and example I needed.

romanzenka:
No problem. I will gladly help with the other tracks as well!

PR #554, opened 2018-03-13
center .igv-search-container correctly

romgrk:
screenshot from 2018-03-13 11-49-16

screenshot from 2018-03-13 11-49-01


jrobinso:
Thanks

PR #585, opened 2018-04-03
Viewport. RulerTrack. SequenceTrack. Remove zoom in notice.

turner:


gungorbudak:
@turner @jrobinso What was the rationale behind removing zoom notice for SequenceTrack? There is always an empty track when you first open a browser and it is not clear what it is until you zoom in. So I think the message is necessary.

jrobinso:
@gungorbudak There is no zoom in notice for the sequence track, are you referring to the gene annotations?

jrobinso:
@gungorbudak Sorry I misread the question and answered the opposite one.   You should not notice the sequence track, it is only a few pixels high,  it has never had a notice by intention in any IGV version (desktop or web).    I don't know what this commit is about, perhaps @turner knows, but if it was there it was a bug.  Ditto the ruler track.

turner:
As Jim stated. By design Ruler and Sequence track do not require a zoom notice. Can you please clarify if this does not answer you question?

gungorbudak:
@jrobinso  @turner I agree that the track is very short but as it has its own controls like the cog button and right context menu, I thought it is not something like RulerTrack but it is a feature track just like others. Based on this thought, I think it can improve UX (at least on me) to treat it like a feature track but if not doing so is a collective decision and the design requirement based on some other considerations, I'll live with it.

I'm just saying with the current design, a new user might not expect to see sequences there and won't zoom in either as it doesn't say so and s/he would have to discover by chance or this will be written in the user guide or something else.

Thanks much for the replies!

jrobinso:
I actually agree that the cog and other ui elements shouldn't be visible when there is no sequence.  In fact the track probably shouldn't be there at all until you are zoomed in enough to see sequence.  However having that text there all the time will annoy most users who aren't looking at the level most of the time.   The current design mirrors the UCSC genome browser, and every other one that I am aware of, with the exception of the visible cog and track dragger.  I do agree that should not be there.

A genome browser without sequence would be a strange thing, but I appreciate your input, we will see what we can do to make this more intuitive.

gungorbudak:
I agree that the notice might be annoying. Thanks for the discussion @jrobinso

jrobinso:
@gungorbudak Thank you for raising the issue, we get used to seeing things like that phantom gear box and forget they are there.  We have discussed it and decided to hide the track when zoomed out.

PR #598, opened 2018-04-05
Additional Track Gear Popup Fixes

turner:


jrobinso:
Doug, why is this incorrect?  It seems essential.

Issue #637, opened 2018-04-19
Display error message when track cannot be loaded


romgrk:
Currently, there is no visual warning if a track could not be loaded (e.g. network error, server error, etc). It would be great if there was a message to let the user know that the track is not loaded, because the current display can mislead the user to think the track is just empty instead of not loaded.

Proposition:
![screenshot from 2018-04-19 12-20-30](https://user-images.githubusercontent.com/1423607/39004883-92dd437a-43cc-11e8-860e-c9974316b9b2.png)

@jrobinso if you're ok with this proposition I can submit a PR to make the changes

jrobinso:
Yes, agreed.  Could you make the PR to the "development" branch?  This allows us to test before committing to master.

PR #639, opened 2018-04-19
Display error message when track cannot be loaded

romgrk:
Fixes https://github.com/igvteam/igv.js/issues/637

Result:
screenshot from 2018-04-19 15-53-19

I'm making this PR on `development` as specified by @jrobinso.

A few suggestions regarding the development process:
 - Maybe use eslint? There are a lot of unused-variables, redefined-variables (declared twice), undefined-variables (probably stuff that is assumed to be global) and unreachable-code. All these can be detected by eslint, here is an example output that I run on this branch: https://gist.github.com/romgrk/e77cf8806b99f005174d6925f6b778e5
- Write a document to describe the development process (how to test manually, how to run the automated tests, how to compile stuff, etc)
 - Remove compiled files from VCS (.css, .css.map)


romgrk:
The "spaced-comment" is a formatting rule in my top-level config to force having a space after `//` (can be safely ignored), but yes, the rest of the rules can catch quite a few bugs.

I'll check if I have time to make a PR for tests, but if I don't have time and you're interested, [GoogleChrome/puppeteer](https://github.com/GoogleChrome/puppeteer) is awesome for these kind of things, matched with any decent testing framework (jest, mocha).

PR #644, opened 2018-04-20
fix str changes after updates

ishanley:
Made some changes so the new STR features would still work in VariantTrack. I also made a change to the trackMenuItem method in trackCore because it was not updated for the new inputDialog and I was using it for the menu in the variantTrack.


jrobinso:
Thanks Issac.

jrobinso:
There are merge conflicts that need to be resolved.   I suspect its the inclusion of the files dist/igv.js,  dist/igv.min.js,  and dist/igv.min.map.    Remove those from your branch and try the pull request again.

PR #696, opened 2018-07-31
Aed format

romanzenka:
Initial implementation of AED format parser.

Needs a lot of work for the byte sources as the AED supports different Unicode encodings. However, the files we see in our practice are simple UTF-8 only with a BOM.

### DIFF
--- a/js/feature/dataWrapper.js
+++ b/js/feature/dataWrapper.js
@@ -59,6 +59,28 @@ var igv = (function (igv) {
         }
     }
 
+    // For use in applications where whitespace carries meaning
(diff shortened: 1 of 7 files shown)


romanzenka:
Sadly, I do not have anything publicly available - we use igv.js internally.

I would like to talk to you someday about what we are attempting to do with this.

When I mentioned AED in igv.js, our users got very excited because they thought AED works also in the desktop version of IGV. I may have to add support there as well. They have been converting AED to BED manually using Excel - no wonder they are excited about direct support!

PR #698, opened 2018-07-31
Making sure that the AED tests run

romanzenka:
I had trouble seeing my tests run, because other failing tests prevented their execution.

- fixed a bug in bwReader
- "fixed" BAM test (does not even attempt operation that is destined to fail)
- fixed bigwig test

### DIFF
--- a/js/bigwig/bwReader.js
+++ b/js/bigwig/bwReader.js
@@ -522,6 +522,14 @@ var igv = (function (igv) {
         var chromToId = {};
         var idToChrom = [];
 
+        this.header = {
+            magic: magic,
+            blockSize: blockSize,
+            keySize: keySize,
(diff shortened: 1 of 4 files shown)


jrobinso:
Thanks,  these are not so much tests now as development tools but they should all run.    How are you running the tests?    I would like to automate them but do not have time to figure out how to do that.   When I run them it is from a web browser with "runTests.html".

romanzenka:
I am running the tests with runTests.html as well, however certain types of errors prevent all tests from running. So in order to run MY tests, I have to disable all failing tests first.

I did experiment with running the tests using grunt, but I am on a newer node and grunt fails for me. I can try to make the tests run using Grunt again - would you be okay with updating the grunt version?

What version of node are you using?

Issue #701, opened 2018-08-05
Module system in NPM package for igv.js 2.0


eweitz:
Following up on the newly available ES6 export (#674), I'd like to propose that the default [igv NPM package](https://www.npmjs.com/package/igv) prioritize support for `import` via ES6 as part of the upcoming igv.js 2.0 release.  

The current package supports `require` and not `import`, and it seems impossible to easily support non-standard and standard module systems in the same NPM package.

I see two options:

1. Support `import` in the default package, and provide a separate `igv-require` package
2. Support `require` in the default package , and provide a separate `igv-es6` package

Neither of those options would break anything, but they would optimize igv differently.

Option 1, making `import` the easier approach, would ensure that more new dependents (and more modern existing dependents) will be able to easily include the default package like they do their other dependencies.  The ES6 `import` standard has already replaced the non-standard `require` as the preferred way to include dependencies in all major JS frameworks and most modern JS libraries. That trend will only increase during the life of igv.js 2.x.

Option 2, making `require` the easier approach, would ensure that the handful of existing dependents using the igv NPM package can continue using the default package as they do now. Although some dependents use igv via NPM and `require`, given embedders I know and my GitHub searches of igv.js usage patterns, I suspect the vast majority include igv.js by hand-written build steps that pull it in by pointing to a URL at build- or run-time. And those few who get igv.js via NPM but do not migrate from the non-standard `require` syntax would still be supported by a separate `igv-require` package.

Either option would work, but I think the case for Option 1 is stronger.

jrobinso:
To summarize,  igv.js will have 2 NPM packages,  one for use as an ES6 module (using "import"),  the second for all other uses  ("require", AMD, and plain script tag imports).   The question is,  which package gets the name "igv",  and which a qualifier.    BTW I'm not sure about "require" as the qualifier as its really everything other than an ES6 module, including "require" and plain script includes.

I don't see a big difference wrt ease-of-use, in both cases you have to find the correct package for the module system your are using.

@vrockai @renyaoxiang @paul-shannon @MariaNattestad  You have all mentioned NPM or require in issues posted here.   If you have an opinion here please post it here.   Also, if you use NPM please post a short note on how you are using it.   Note that the proposed change is for the upcoming 2.0 release.

vrockai:
[...]

PR #702, opened 2018-08-07
Provide easily usable ES6 export, improve package metadata

eweitz:
This enables including igv.js using ES6 `import` as done in modern React, Angular and Vue, while retaining support for AMD `require` and plain `script` tags, as discussed in #701.  

It also adds `scripts` and `files` properties to package.json to build igv.js artifacts before publishing to NPM, and to strip other files from the package.  And it puts metadata previously only visible in the published NPM package into source control, to be more open source and in line with mainstream JS conventions.

See the following for examples of various ways to include igv.js, using the experimental [tmp_es6_igv](https://www.npmjs.com/package/tmp_es6_igv) NPM package:
* [ES6 `import` in React](https://github.com/eweitz/igv.js-react), specifically [package.json](https://github.com/eweitz/igv.js-react/blob/5f95ca7f4ec3b9a67820fa87f8b34d6548cb0cd7/package.json#L6) and [App.js](https://github.com/eweitz/igv.js-react/blob/master/src/App.js#L6)
* [AMD `require`](https://jsfiddle.net/mogbkej0/5/)
* [`script` tag](https://jsfiddle.net/mq54x2y6/6/)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,33 @@
 {
-  "name": "IGV",
-  "version": "0.0.1",
+  "name": "igv",
+  "version": "2.0.0-rc5",
+  "description": "Embeddable genomic visualization component based on the Integrative Genomics Viewer",
+  "main": "dist/igv.esm.js",
+  "files": ["dist/**"],
+  "scripts": {
+    "prepublishOnly": "grunt"
+  },
+  "author": {
+    "name": "Jim Robinson",
+    "email": "igv-team@broadinstitute.org",
+    "url": "https://igv.org"
+  },
+  "bugs": {
+    "url": "https://github.com/igvteam/igv.js/issues"
+  },
+  "bundleDependencies": false,
+  "deprecated": false,
+  "homepage": "https://igv.org",
+  "keywords": [
+    "IGV",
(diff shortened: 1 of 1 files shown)


jrobinso:
Looks good @eweitz -- but could you comment on the impact, of any of using "main" instead of "browser"?  The NPM documentation says that "browser" should be used instead of main if the module is intended to client-side use.   Is this just a matter of style?  https://docs.npmjs.com/files/package.json

eweitz:
> the impact, if any of using "main" instead of "browser"

I don't foresee any impact in the near- or medium-term.  It seems that `browser` in package.json might better enable bundling for both Node apps and web apps, and perhaps better tree-shaking.

I'm inclined to stick with `main` and ignore `browser` for a few reasons: 1) I don't imagine Node support or aggressive tree shaking are near-term goals for igv.js, 2) I haven't used the `browser` field, and 3) ES6 `import` in [igv.js-react](https://github.com/eweitz/igv.js-react) works fine with `main`.

jrobinso:
OK, merged, thanks much!

PR #706, opened 2018-08-09
Enable previewing NPM package via `npm pack`

eweitz:
The [package.json improvements](https://github.com/igvteam/igv.js/commit/0c551bb1dcfe3f062671d83ec055faf06acf471c) in #702 included adding `scripts` and `files` properties to package.json to build igv.js artifacts before publishing to NPM, and to strip other files from the package.  

While replying to [Jim's comment from yesterday](https://github.com/igvteam/igv.js/issues/701#issuecomment-411629051), it occurred to me that the `scripts` property could be refined.

This updates the `scripts` property to run Grunt during `npm pack`, in addition to `npm publish`.  It enables developers to preview what the igv NPM package will contain before publishing.

Using package.json has benefits over using external (e.g. Jenkins) build scripts: 1) it is open source (with all the benefits that brings), 2) it's in line with mainstream JS conventions, and 3) it requires less maintenance of things handled out-of-the-box by the `npm` client.

Here's an example of how this would be used:
```
# Ensure devDependencies like grunt are available

$ npm install --only=dev
npm notice created a lockfile as package-lock.json. You should commit this file.
added 130 packages from 104 contributors and audited 303 packages in 2.778s
found 0 vulnerabilities

# `npm pack` enables you to preview the tarballed package that goes to NPM.
# The files that go into the tarball are defined by the `files` property in package.json, 
# and by default includes conventional files like any README and LICENSE.
# `npm pack` gives a dry run of `npm publish`.

$ npm pack

> igv@2.0.0-rc5 prepack /Library/WebServer/Documents/igv.js
> grunt

Running "concat:css" (concat) task

Running "embed-css" task

Running "concat:igv" (concat) task

Running "concat:igv_es6" (concat) task

Running "uglify:igv" (uglify) task
>> 1 sourcemap created.
>> 1 file created 2.45 MB → 1.48 MB

Running "uglify:igv_esm" (uglify) task
>> 1 sourcemap created.
>> 1 file created 2.45 MB → 1.47 MB

Done.
npm notice 
npm notice 📦  igv@2.0.0-rc5
npm notice === Tarball Contents === 
npm notice 843B    package.json           
npm notice 1.1kB   license.txt            
npm notice 295B    README.md              
npm notice 2.5MB   dist/igv.esm.js        
npm notice 1.5MB   dist/igv.esm.min.js    
npm notice 799.7kB dist/igv.esm.min.js.map
npm notice 2.5MB   dist/igv.js            
npm notice 1.5MB   dist/igv.min.js        
npm notice 799.5kB dist/igv.min.js.map    
npm notice === Tarball Details === 
npm notice name:          igv                                     
npm notice version:       2.0.0-rc5                               
npm notice filename:      igv-2.0.0-rc5.tgz                       
npm notice package size:  2.6 MB                                  
npm notice unpacked size: 9.5 MB                                  
npm notice shasum:        d67aaccc6f2adb703d711177c9b109c77ac82916
npm notice integrity:     sha512-VDuaLP5v/TgEs[...]R7gvmqCHceRQw==
npm notice total files:   9                                       
npm notice 
igv-2.0.0-rc5.tgz
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,11 @@
   "version": "2.0.0-rc5",
   "description": "Embeddable genomic visualization component based on the Integrative Genomics Viewer",
   "main": "dist/igv.esm.js",
-  "files": ["dist/**"],
+  "files": [
+    "dist/**"
+  ],
   "scripts": {
-    "prepublishOnly": "grunt"
+    "prepack": "grunt"
   },
   "author": {
     "name": "Jim Robinson",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,33 @@
 {
-  "name": "IGV",
-  "version": "0.0.1",
+  "name": "igv",
+  "version": "2.0.0-rc5",
+  "description": "Embeddable genomic visualization component based on the Integrative Genomics Viewer",
+  "main": "dist/igv.esm.js",
+  "files": ["dist/**"],
+  "scripts": {
+    "prepublishOnly": "grunt"
+  },
+  "author": {
+    "name": "Jim Robinson",
+    "email": "igv-team@broadinstitute.org",
+    "url": "https://igv.org"
+  },
+  "bugs": {
+    "url": "https://github.com/igvteam/igv.js/issues"
+  },
+  "bundleDependencies": false,
+  "deprecated": false,
+  "homepage": "https://igv.org",
+  "keywords": [
+    "IGV",
+    "genomics",
+    "visualization"
+  ],
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/igvteam/igv.js.git"
+  },
   "devDependencies": {
     "grunt": "1.0.3",
     "grunt-contrib-concat": "1.0.1",


jrobinso:
This looks great.   Have you been able to test the packaged npm module?

jrobinso:
I don't like the hardocded version number in the package.json,  it would be very easy to publish some random commit from master as the version specified.

jrobinso:
I can't seem to run npm pack.  I get the error below.  Clearly I have grunt-cli installed, I can run grunt from the command line.   Do I need to add grunt-cli to "dev-dependencies"?
```
hlbrfiixx2a3201:igv.js jrobinso$ npm pack

> igv@2.0.0-rc5 prepack /Users/jrobinso/Dropbox/projects/igv.js
> grunt

internal/modules/cjs/loader.js:583
    throw err;
    ^

Error: Cannot find module 'grunt-cli/bin/grunt'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
```

eweitz:
> I don't like the hardocded version number in the package.json, it would be very easy to publish some random commit from master as the version specified.

[`npm version`](https://docs.npmjs.com/cli/version) addresses that.  When you run `npm version 2.0.0-rc5`, it updates the "version" field in package.json then commits to your local Git repo.

Then you'd run `npm publish` (or `npm pack` to preview) to publish the package to NPM and `git push` to publish the code to GitHub.

You're right that having a future version specified in package.json well in advance of publishing to NPM isn't a best practice.  (My bad!)  But the standard practice outlined above essentially eliminates the possibility of accidentally publishing, while also ensuring that all package metadata is under open source control, per convention.

> Clearly I have grunt-cli installed, I can run grunt from the command line. Do I need to add grunt-cli to "dev-dependencies"?

Good point, I suspect so.  I have `grunt-cli` installed globally per [its default installation instructions](https://github.com/gruntjs/grunt-cli#grunt-cli--).  But they mention [it can be installed locally](https://github.com/gruntjs/grunt-cli#installing-grunt-cli-locally) via `npm install grunt-cli --save-dev`, which adds it to `devDependencies`.  

(I don't think we need to revise `scripts` in package.json as those grunt-cli docs note, because running igv.js tests via CLI is basically not supported.)

jrobinso:
If you try to publish with a previously published version number will NPM reject it with an error?

eweitz:
Yes, it will.

jrobinso:
OK, good, then yes the procedure outlined is safe.   So the general rule would be the version number should lag behind the state of master.  This will mean the version in package.json does not reflect the current state of the git repository,  but I guess this doesn't matter unless someone is trying to install directly from git.

eweitz:
Great.  And you're right that the version number in package.json would lag behind the state of master (except for the one commit marking the package publication), which is conventional and perfectly fine.

PR #722, opened 2018-08-21
Make sure NPM works, yet prevent accidental publishing of the package.

romanzenka:
Same stop-gap measure as before - prevent publishing - but this does not break npm because of invalid version string.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "igv",
-  "version": "VERSION",
+  "version": "2.0.0-rc5",
+  "private": true,
   "description": "Embeddable genomic visualization component based on the Integrative Genomics Viewer",
   "main": "dist/igv.esm.js",
   "files": [


jrobinso:
I don't understand the consequence of marking this "private",  but I suspect that's not something we want to do.   Secondly,  I don't see how we "break npm" since we haven't pushed to NPM.

After the next tag (rc5) this will cease to be an issue,  the "VERSION" string is there to prevent an accident before that time.    We will be pushing that tag soon,  within a day or two.

eweitz:
@romanzenka, see https://github.com/igvteam/igv.js/pull/706#issuecomment-412099158 for context on Jim's comment.  

Seeing `"version": "VERSION"` in package.json was concerning to me as well, but it sounds like igv.js will be using the more familiar practice of setting `version` in the package.json on GitHub to the current published NPM package version as of 2.0.0-rc5.  Accidental publishing will then be a non-issue, as NPM will refuse to overwrite an existing package version.

jrobinso:
Yes, the VERSION is temporary as I have nothing else to put there until we
do a push with the new process.   That will be soon.    Its VERSION on
purpose so publishing won't work until that time,  NPM would reject it,
 not that an accident is likely,

jrobinso:
Solved in a different way, we will keep the last version number published in package.json.    I think this is the general convention.

PR #723, opened 2018-08-21
Avoid save/restore whenever possible

romanzenka:
The save/restore pairs are quite slow especially in loops executing
thousands of times. Usually the save/restore is needed only when
properties are being set. For untransformed text drawing, a fast
path was added that draws text directly without save/transform/restore.


jrobinso:
Very nice.  I should get a slower computer so I notice these things.

PR #724, opened 2018-08-21
Remove second removeBrowser function

Chris7:
There are 2 removeBrowser functions in igv-create.js, with the second one throwing an error upon usage. This function is needed as we are using virtual lists to render thousands of sequencing results and need to cleanup the browsers that are no longer in the view. (Though once we scroll through about 10 browsers, it still crashes)

### DIFF
--- a/js/igv-create.js
+++ b/js/igv-create.js
@@ -471,12 +471,6 @@ var igv = (function (igv) {
 
     }
 
-    igv.removeBrowser = function (browser) {
-        browser.$root.remove();
-        browser.dispose();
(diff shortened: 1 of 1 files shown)


jrobinso:
You're using a function that hasn't been tested at all really,  hasn't been documented,  and might not even exist in its current form tomorrow.     There should not be 2 versions obviously,  but the first version (the 1 kept) is not complete.   I suggest you wait before trying to use this function.

Chris7:
Thanks, I made an issue here https://github.com/igvteam/igv.js/issues/726 with a demo app.

PR #736, opened 2018-09-18
New features

renkyp:
Add new features include:

groupby tag option
show softclip 
add locus info with cursor guide
mouse wheel move inside track view 
drag bar for each track view


jrobinso:
@renkyp Thanks for the contribution!  Sorry for the delay in commenting, we're working furiously on a final 2.0 release and I have been traveling to boot.   There's a lot of good stuff here,  but I won't be able to review it until we get this release out.   Hopefully late next week.    Also, its very hard to review and accept a large pull request with multiple unrelated features like this,  if you can break it up that would be helpful.   Unfortunately its not possible to accept pieces,  its all or nothing apparently.

PR #744, opened 2018-10-04
issue743fix : fix to select correct alignmentBlock when the clicked p…

kaushik51094:
fix to select correct alignmentBlock when the clicked position is right after an insertion


jrobinso:
You're solution is correct, coordinates are exclusive of end.  Thanks for the PR.

PR #754, opened 2018-10-16
Resize trackView left gutter properly

anderspitman:


anderspitman:
Hey @jrobinso this probably shouldn't be merged yet, since it's an incomplete solution, but I wanted to show how I'm fixing it for the left gutter.

anderspitman:
I think we need to do the same for the reference sequence and the loading icons (don't actually know where those live yet). Possibly other pieces as well.

jrobinso:
I don't understand the problem yet, or how the solution helps, but will look into it.

anderspitman:
Basically the sub-components aren't completely resizing themselves when the IGV browser resizes.

jrobinso:
Its interesting,  I'm watching the elements with developer tools (tabs example).   When the root container becomes visible some of the elements resize immediately, and in fact they all do eventually including the content canvas.    It just takes 10-15 seconds!   Its possible developer tools is triggering the resize.   I think you're fix triggers the resize, but perhaps there's something more direct we can call to do it.   Investigating.
Its also interesting that the gene track works right away.    Why should this behavior be track type specific?

anderspitman:
Oh, that is very interesting. I tried waiting and I'm never seeing them, both in Firefox and Chrome. Here's my modified initIGV from tabs.html:

```javascript
function initIGV() {

        var igvDiv, options;
        igvDiv = document.getElementById("igvDiv");
        options =
            {
                genome: "hg19",
                locus: "chr8:128,747,267-128,754,546",
                tracks: [
                    {
                        type: 'alignment',
                        format: 'bam',
                        url: 'https://data.broadinstitute.org/igvdata/1KG/b37/data/HG02450/alignment/HG02450.mapped.ILLUMINA.bwa.ACB.low_coverage.20120522.bam',
                        name: 'HG02450'
                    }

              ]
            };
        igv.createBrowser(igvDiv, options);
    }
```

jrobinso:
OK, progress.   The reason for the delay, and the reason the gene track works right away,  is you don't see anything until the data is loaded for the view.   Normally you should see the spinner while its loading.

anderspitman:
Oh so yours wasn't loading anything for 10-15 seconds? Normally igv.js loads the data in 1-3 seconds for me.

jrobinso:
It depends on your server, network, etc,  and 10-15 is exaggerated.   It seemed like that but was really 4-5.

I just pushed a fix for the tabs case.   There were 2 issues.   (1) an "updateViews" call was needed in the resize() function, primarily to kick off the spinner.   (2) the spinner was invisible.   This was because it was initialized with a size == min(content height, 32).    In this case content height was zero initially.   Its now set to a constant 32.    This fixes the tab-with-bam case for me,  I don't have a modal test case set up but I think it should fix that as well.   Let me know.

anderspitman:
Excellent, the spinners are working for me now. However, some of the axes and the sequence are still missing on my machine. Here's what I'm seeing:

untitled drawing

jrobinso:
@anderspitman I just pushed fixes for the ruler & sequence tracks,  will look at the axis later.

anderspitman:
Seems to be working for the tabs demo. FWIW this pull request affects the coverage y-axis. Not sure if it's an appropriate final fix, but might help you drill down on the root cause. Have fun at ASHG. A bunch of our lab is down there as well.

jrobinso:
@anderspitman -- your PR was the correct solution for the axis, thanks!  merged.

anderspitman:
Awesome, glad it ended up working out. The tabs demo appears to be working correctly for me now. Thanks again for your help!

PR #755, opened 2018-10-17
Responsive navbar

turner:
Responsive navbar


turner:
@jrobinso the conflict has to do with removal of the browser.js responsive code. Just remove that code and it should merge just fine.

PR #774, opened 2018-11-02
Correct rounding error in bam coverage calculation

oyvinev:
Fix rounding error in bam coverage, where Math.floor was used without correcting the input.


jrobinso:
Thanks for the PR.  Would Math.round() accomplish the same thing?

jrobinso:
Thanks for contributing.

PR #790, opened 2018-11-08
Show genome name

matthdsm:
Shows the genome "name" in the browser header instead of the id.
Tested locally.
The spacing between the "IGV" logo and the genome name could use some work though.
If you don't like this, feel free to close

Cheers
M


jrobinso:
For our own app (igv.org/app) we wanted that name to be short, to take minimal space on the toolbar, so we use ID.    We anticipate space on the toolbar to get increasingly tight as more buttons are added.  Would it suffice to make the full name appear when hovering over the ID?

PR #832, opened 2019-01-11
Declare stainColor

anderspitman:


jrobinso:
Thanks.  I'm going to start a code modernization effort soon including lint that should catch all these.

anderspitman:
No problem. This one started throwing an error for me. Which is weird because it's been there for a while and I hadn't thought I changed anything that should affect it.

PR #876, opened 2019-03-19
show strand marks on all features

SchwarzMarek:
Adds strand/direction (>/<) mark to all features.
Reverse strand must be explicitly specified. (previous implementation adds reverse if strand is not "+")
If nothing is specified no strand mark added.

### DIFF
--- a/js/feature/featureTrack.js
+++ b/js/feature/featureTrack.js
@@ -509,17 +509,31 @@ var igv = (function (igv) {
 
             const exonCount = feature.exons ? feature.exons.length : 0;
             const coord = calculateFeatureCoordinates(feature, bpStart, xScale);
+            const step = this.arrowSpacing;
+            const direction = feature.strand === '+' ? 1 : feature.strand === '-' ? -1 : 0;
+
             if (exonCount === 0) {
                 // single-exon transcript
                 ctx.fillRect(coord.px, py, coord.pw, h);
-
+                
+                // Arrows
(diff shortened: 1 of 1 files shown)


jrobinso:
Hi, thanks for the PR.   Could you attach a screenshot?   IGV and igv.js draws strand on the introns for multi-exonic transcripts.   It has never drawn them on the exons.   To do so would be a major change on 10+ years of convention.   Also we want to maintain the same look as UCSC.

SchwarzMarek:
Hi,
This would not break the convention, as nothing is changed in multiexonic transcript.

I've came across this with bacterial genomes where most CDS are without introns. 

The default IGV shows picture like this:
Screenshot_strands_off

You cannot tell on which strand the gene is encoded.

The PR would result in this:
Screenshot_strands_on

Here you can see on which strand the gene is, which is valuable information.

Maybe I should add some switch to track configuration which would enable this? But as I've said, no change in appearance for multi-exonic transcripts is implemented (apart form fix for case when no strand is specified for the feature)

Best regards

jrobinso:
@SchwarzMarek Ah, thanks for the clarification.   Yes this is a good addition, in fact it is the intended behavior so this is a bug fix.   Have you tested it on a transcript with introns?

SchwarzMarek:
Hi,
I don't have anything with multiple exons right now, so I've used the cram.html in your test directory.

default igv (current master)
igv_default

proposed:
igv_proposed

Best regards

PR #880, opened 2019-03-27
Bug fix preventing use in tracing3D.

turner:
Bug fix preventing use in tracing3D.


jrobinso:
I see a bunch of format differences but where it the bug?  Did you format with intellij before pushing?

PR #936, opened 2019-07-10
Update gtex parser

dnguyen1193:


jrobinso:
Hi, what is the difference in "v7" and "v8".   Is v7 for hg19, and v8 for hg38?   Will hg19 continue to be supported?

dnguyen1193:
Yes, v8 will be in hg38. We will not be supporting hg19 anymore, but will have a mapping from the hg38 to hg19 position. v8 will also contain more data.

PR #942, opened 2019-07-31
Minor refactors to igvxhr.js

adamkewley:
Hi, 

I'm working on an integration with igv.js and, while reading through `igvxhr.js`, refactored some of the code as I went along. Most changes are just adding semicolons, changing control flow to remove indentation (notably: when a function has an early assert + return check at start of a function) and removal of C-style variable forward declaration.

igv.js is excellent, these are just costmetic refactors that the developers might like to merge in. No biggie if you're busy with other things.


jrobinso:
Yes, this is good, merged.     I'm cleaning up slowly.   The c-style forward declaration for var was to make clear that the variable is function scope,  from many years ago,  these are being replaced over time with const and let and forward declaration removed then.  Your PR saves me a little time.

RE semicolons no final decision but they might completely disappear at some point,  but agreed the style shouldn't be mixed.

PR #945, opened 2019-08-01
Make all normal dependencies dev dependencies

zarsky-broad:
This moves all normal dependencies to dev dependencies, since they are unused when importing IGV as a dependency. Of note, `@gmod/bam` and `shelljs` can be removed entirely without effecting anything, but that was outside the scope of this.

The reason for wanting to do this was the warnings being generated when I installed IGV as a dependency to my project, which doesn't use `grunt`:
```
npm WARN grunt-contrib-clean@2.0.0 requires a peer of grunt@>=0.4.5 but none is installed. You must install peer dependencies yourself.
npm WARN grunt-contrib-connect@2.0.0 requires a peer of grunt@>=0.4.0 but none is installed. You must install peer dependencies yourself.
```

This has no effect on the built output - aside from marking packages as `dev` in the package-lock, the only changes are the same as those observed when running `npm install` and `npm run grunt` against master.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "igv",
-  "version": "2.2.13",
+  "version": "2.2.15",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


jrobinso:
Yes, completely agree, and also gmod/bam is not used as you say.   I'll remove it.  Thanks for the PR

PR #977, opened 2019-09-17
Fixed edge-case bug where the least significant byte of bsize can overflow

adamkewley:
Encountered an issue where igv.js would request an invalid block size from the API followed by the ZLib library whining about the input buffer being too small compared to the size header in the data (uncompressed DEFLATE block). This fixed that issue, which is because the existing code would overflow `ba[16]` if it was equal to `255`.

To test this in isolation, open a browser console and try:

```js
const a = new Uint8Array(2);
a[0] = 255;
a[1] = 255;
// == 65280, which is incorrect
const broke = (a[1] << 8) | (a[0]) + 1;

// == 65536, which is correct
const fixed = ((a[1] << 8) | (a[0])) + 1;
```


jrobinso:
wow, great catch. thanks.

PR #979, opened 2019-09-19
Optimized BAI 'optimizeChunks'

adamkewley:
This patch contains a possible optimization to the BAI layer of igv.js. 

The significance of the BAI layer is that it figures out what HTTP requests are necessary to display a basepair range in the UI. The [BAI spec](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&ved=2ahUKEwism6nZmN3kAhVhQxUIHdPxDzoQFjAAegQIAhAC&url=https%3A%2F%2Fsamtools.github.io%2Fhts-specs%2FSAMv1.pdf&usg=AOvVaw0MUgu2LimUquQxdPowaiZp) explains how a BAI file can be used to calculate which chunks in an rtree *may* overlap with the queried basepair range. Many of those chunks do not overlap with the query range at all, though, because they are chunks that appear in upper bins that the algorithm can't logically eliminate with the linear index because they lie to the right of the query range.

To avoid reading chunks to the right of the query range, the spec also recommends:

> Given the set of bins that overlap the specified region, we can visit alignments in the order of their leftmost coordinates and stop seeking the rest when an alignment falls outside the required region.

However, because igv.js asks for all chunks in parallel, rather than asking for each sequentially (I guess for latency-related reasons?), it will redundantly fetch data that's nowhere near the viewed range, only to discard it later (after getting it, unpacking it, etc.). 

I implemented an additional optimization pass into the BAI layer (this PR). The optimization pass attempts to figure out what chunks definitely aren't in the query range by assuming that only so many basepairs could fit into a chunk *and* that chunk was assigned to a particular non-leaf bin because it intersected at the boundary between two of the bin's children. If those two assumptions are correct, an algorithm can figure out whether a chunk could intersect with the query range. If it definitely can't, the chunk is dropped from the fetch list. **This optimization results in ~20-80 % fewer fetches when viewing BAM files in igv.js in my test cases**.

I understand that the optimization might appear a little questionable. Maybe I've misunderstood the BAI spec, or my logic is off. I've tried to carefully comment the logic at each step and I'd really appreciate if a dev could confirm that this is a feasible optimization: it *significantly* improves the user-experience in my particular use-case.

Aside: It might be an idea, e.g. as a user-facing option, to enable loading chunks sequentially as recommended by the spec. However, I'm not experienced enough with the codebase to implement that, so I have tried to optimize in a very localized part of the source: this change only affects an internal function in one source file.

### DIFF
--- a/js/bam/bamIndex.js
+++ b/js/bam/bamIndex.js
@@ -178,12 +178,12 @@ BamIndex.prototype.blocksForRange = function (refId, min, max) {
             }
         }
 
-        return optimizeChunks(chunks, lowest);
+        return optimizeChunks(chunks, lowest, min, max);
     }
 
 };
 
-function optimizeChunks(chunks, lowest) {
+function optimizeChunks(chunks, lowest, min, max) {
 
     const mergedChunks = []
     let lastChunk = null
@@ -201,7 +201,7 @@ function optimizeChunks(chunks, lowest) {
 
     chunks.forEach(function (chunk) {
 
-        if (!lowest || chunk.maxv.isGreaterThan(lowest)) {
+        if (!lowest || chunk.maxv.isGreaterThan(lowest) && chunkCouldOverlapWithRange(chunk, min, max)) {
             if (lastChunk === null) {
                 mergedChunks.push(chunk);
                 lastChunk = chunk;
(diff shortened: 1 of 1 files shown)


js/bam/bamIndex.js
```diff
@@ -221,6 +221,102 @@ function optimizeChunks(chunks, lowest) {
...
+    // *could* overlap with the query range.
+    if (chunk.minv.block !== chunk.maxv.block) {
+        return true;
+    }
+
+    // If the chunk is in a "leaf" bin of the rtree, we
+    // pessimistically assume that the chunk *could* overlap with
+    // the query range (it's a leaf bin, so it *probably* does
+    // intersect the query range anyway).
+    if (4681 <= chunk.bin && chunk.bin <= 37448) {
```
adamkewley, line 279: These magic numbers could be renamed to `LEAFBINID_MIN` `LEAFBINID_MAX`

js/bam/bamIndex.js
```diff
@@ -221,6 +221,102 @@ function optimizeChunks(chunks, lowest) {
...
+        return Math.pow(2, 14);
+    }
+}
+
+/**
+ * Returns the basepairs where a bin's children intersect
+ */
+function bpsWhereChildrenIntersect(bin) {
+    const childSpan = binSpan(bin) / Math.pow(2, 3);
+    const numBinsAtLvl = Math.pow(2, 29) / childSpan;
```
adamkewley, line 248: 2^29 maybe should be called `BAI_MAX_BP_RANGE` or something

js/bam/bamIndex.js
jrobinso: OK, thanks for this.  I'm not aware of any good reason for loading the chunks in parallel, if I changed it to load sequentially could we do an optimization closer to the spec recommendation?

jrobinso:
OK, thanks for this. I'm not aware of any good reason for loading the chunks in parallel, if I changed it to load sequentially could we do an optimization closer to the spec recommendation?

js/bam/bamIndex.js
```diff
@@ -221,6 +221,102 @@ function optimizeChunks(chunks, lowest) {
...
+    // we can compute the uncompressed span of the chunk
+    const uncompressedSize = chunk.maxv.offset - chunk.minv.offset;
+
+    // Assuming *all* of the data in the span is, in fact, purely
+    // 4-bit basepairs with no quality scores, tag names,
+    // etc. then the theoretical maximum number of base pairs the
+    // chunk may contain can be calculated from its size.
+    const BPS_IN_BYTE = 2;
+    const maxBpsInChunk = BPS_IN_BYTE * uncompressedSize;
+
```
jrobinso, line 293: I'm not following this logic.  The contents of a bgzf block could consist entirely of alignments at the same genomic location.  Alternatively it could contain alignments that span the entire chromosome.   So I don't understand this 4-bit assumption.  I don't see anyway to determine the maximum # of bps a BGZF block can span without loading and looking at the data.

jrobinso:
@adamkewley Thank you for calling attention to this.  See my alternate optimization in the  branch bai_optimization.   I think this is doing the right thing, but its been a long time since I had my head around this and would appreciate your review.

js/bam/bamIndex.js
adamkewley: So the logic is that we know how many bytes of uncompressed data that particular chunk represents because we only look at chunks that are within one BGZF block (chunks that span multiple blocks are discarded because compression efficiency can't be assumed).

Lets assume the chunk spans 128 uncompressed BAM bytes. That chunk, in the worst possible case, might contain only BAM basepairs (this isn't necessarily possible, but lets be overly pessimistic). Those base pairs, in the worst case, might be encoded with maximum efficiency (4 bits per base pair). Therefore, in this worst case, the BAM data would span a range of 256 base-pairs (unless the uncompressed BAM data format contains a mechanism for encoding a larger range than that in 128 bytes). In reality, because the data may contain multiple alignments, quality scores, etc. then the basepair span of the data is going to be lower than this bound but this logic to specifically establish an upper bound.

From that logic, I am asserting that the maximum "span" of that chunk is 256 base pairs. Asserting that upper bound is a key step in this optimization. The optimization algorithm uses the chunk's bin number to infer all possible positions (in base pair space) that the chunk may have spanned when it was being indexed. If the chunk is in a leaf node of the rtree, all we can say is that "the chunk represents an upto 256 bp range within the range of that bin", which isn't useful because earlier steps in the process have already established that the leaf bin is pertinent to the query (`reg2bins`). 

However, if the chunk was assigned  to an internal bin of the tree--and presuming the chunk was assigned to that node *because* it couldn't fit into a smaller bin--then additional information can be inferred. We know, in this particular case, that:

- The chunk is up to some 256 bp range within the bin (same as a leaf)
- And the chunk must have intersected the boundary between two children bins (hence its promotion into an upper bin)
- So, at worst, there are 7 possible ranges in the bin, 512 bps across and centred on the boundaries between direct descendants in the tree, where the chunk *must* have intersected
- And if our query range does not intersect with any of those 7 ranges, then this chunk cannot contain data within our query range

# Key Assumptions

If these assumptions are false, this optimization breaks

- The maximum density of basepairs in uncompressed BAM data is 2 bps/byte
- Chunks in the BAI file are *always* assigned to the smallest possible bin

adamkewley:
I'll give your branch a try. I think that, if it does what the BAI spec says (read each chunk sequentially then quit) then your solution will be *much* more robust than this optimization and that this optimization idea should be dumped if sequentially reading blocks is fine.

This optimization explicitly relies on a logically correct BAI structure that follows the spec's recommendations (which are not explicitly *strict* in saying that implementations MUST assign bins optimally). It will break if an indexing algorithm decided to (sub-optimally) dump all chunks into (e.g.) bin 0.

I wrote this optimization because over-fetching data is *extremely* expensive in my integration of igv.js, not because of latency, but because of throughput limitations. I wasn't sure whether igv.js is used in scenarios where latency is a killer, and didn't know how to bend the codebase accordingly, so I tried to implement something that doesn't affect that part of the system.

js/bam/bamIndex.js
adamkewley: Ah, @jrobinso , you're right in that the chunk could contain multiple reads that are aligned to very different parts of the genome, so a counterexample would a chunk containing two reads on opposite ends of the genome, and that chunk being the only chunk in the file (for example).

adamkewley:
Checked out `bai_optimization` and tried it for our use-case, where it reduces requests by ~60-90 % for the particular dataset I'm working with :+1:

adamkewley:
Also, we managed to find counterexamples to the way this particular optimization works. Imagine there is a BAM file containing only two reads. One of those reads is aligned to the start of a reference, the other read is aligned to the end. This algorithm would assume the maximal span of that chunk can be found from the data size of the chunk but, in this case, it can't.

So, in summary, the optimization I presented is a dud, but @jrobinso 's optimization is a robust improvement that I hope gets mainlined.

jrobinso:
@adamkewley OK, thanks again for raising this issue, and all the testing.  I will close this pull request and merge the bai_optimization branch.

PR #1011, opened 2019-10-28
bug fix - await on igvxhr.loadString

dtsamitros:
it seems that the `await` operator is missing since `igvxhr.loadString` is async..


jrobinso:
You are correct!  We don't use tribble indexes here, so this was not noticed.  Thanks for the PR.

PR #1015, opened 2019-11-07
AED allows a 0 for feature start.

romanzenka:
Unfortunately, 0 is false-y so the test was incorrectly failing to parse a legitimate feature.

Sorry, this came out only when I was doing a very thorough test, normally people have no features at 0.


jrobinso:
No problem, that one has bitten me on several occasions also.

Issue #1023, opened 2019-11-20
Add support to draw horizontal line in bg/bw tracks


blajoie:
Hi Jim,

We (@docwarren) would like to add functionality to draw horizontal 'guide' lines on any given track (mainly bg,bw).  See attached screenshot.

Do you think such a feature would have wide appeal?  If so happy to PR.

To implement, we were considering adding a new track param (halign) that can take in an array of values where a horizontal line will be drawn across the track.  Ideally one could also control the line-type (solid, dotted etc) and even the color.  

Param input could then be something like a object'
halign: [{"y":100,"color":"blue","line_type":"dotted"},{"y":200,"color":"red","line_type":"dotted"}]

[](url)
![horizontal_line](https://user-images.githubusercontent.com/8889420/69264337-f951ec00-0b7b-11ea-93e2-5f548d491b32.png)

jrobinso:
This looks reasonable, and useful, but for line_type check to see what canvas options there are.  I'm not sure myself, possibly you would find this in strokeStyle.

blajoie:
Will do - more soon @jrobinso

docwarren:
https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setLineDash

blajoie:
> https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setLineDash
> @blajoie I don't think we should try to do this for all tracks. Its only relevant for tracks that have a numerical y axis.

Yes exactly, this is only logical for 'data' tracks.  bedgraph/wig

docwarren:
@blajoie We will need to update the wiki if these changes are accepted

jrobinso:
This has been merged

PR #1033, opened 2019-12-10
Hor lines

docwarren:
See #1023

### DIFF
--- a/js/feature/wigTrack.js
+++ b/js/feature/wigTrack.js
@@ -35,6 +35,7 @@ import {createCheckbox} from "../igv-icons.js";
 import {numberFormatter} from "../util/stringUtils.js";
 import {extend} from "../util/igvUtils.js";
 import FeatureTrack from "./featureTrack";
+import {visibilityChange} from "../igv-create";
 
 const dataRangeMenuItem = MenuUtils.dataRangeMenuItem;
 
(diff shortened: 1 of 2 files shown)


js/feature/wigTrack.js
```diff
@@ -137,6 +138,66 @@ WigTrack.prototype.draw = function (options) {
         baselineColor = IGVColor.addAlpha(self.color, 0.1);
     }
 
+    const drawGuideLines = function (options) {
+        if (self.config.hasOwnProperty('guideLines')) {
+            for (let line of self.config.guideLines) {
+                if (line.hasOwnProperty('color') && line.hasOwnProperty('y') && line.hasOwnProperty('dotted')) {
+                    options.context.setLineDash([5,5]);
```
docwarren: I need to change this so that it only sets line dash if the line.dotted property is true;

docwarren: Fixed in later commit

js/feature/wigTrack.js
```diff
@@ -142,18 +143,13 @@ WigTrack.prototype.draw = function (options) {
...
-                            'strokeWidth': 2
-                        }
-                        );
-                    options.context.setLineDash([]);
+                    let y = getY(line.y);
+                    let props = {
+                        'strokeStyle': line['color'],
+                        'strokeWidth': 2
+                    };
+                    if (line['dotted']) IGVGraphics.dashedLine(options.context, 0, y, options.pixelWidth, y, props);
```
docwarren: need to pass dash length - fixed in next commit

Issue #1013, opened 2019-10-31
API to remove track and remove ROI


pratibha-strand:
We have APIs to loadTrack and loadROI. It will be good to expose APIs for removing them as well.

jayfmQQ:
In the latest version(igv@2.3.3),  you can use `browser.removeTrackByName(name)` to remove a exists track.

pratibha-strand:
@jayfmQQ Thanks for the update. But we still do not have an API to remove ROI.

pratibha-strand:
@jrobinso Can you include API to removeROI in the next release. Thanks.

jrobinso:
Thanks @pratibha-strand for the PR.   I'm leaving this open until we release and update the wiki doc.  Also - note to self - we should add a similar method to BaseTrack for the track-specific ROIs.

PR #1039, opened 2019-12-13
Adding API to removeROI

pratibha-strand:
Check #1013


jrobinso:
Great!  Thanks a lot.

Issue #1018, opened 2019-11-11
Read track description from config


pratibha-strand:
There is no way to set track description from config. It will be good to read and set this from the config.

PR #1042, opened 2019-12-17
Adding ability to set custom track description

pratibha-strand:
Check https://github.com/igvteam/igv.js/issues/1018


pratibha-strand:
check https://github.com/igvteam/igv.js/issues/1018

Issue #1049, opened 2020-01-24
ERROR: range-byte header was ignored for url


romanzenka:
This is basically the same issue as #680. 

We are doing some testing, so we are loading very small files. 

Chrome seems to give up asking for tiny file slivers, instead delivers entire file contents all at once.

IGV.js considers this a serious error and reports it.

Technically it is not an issue that you got more data than what you asked for! 

I will try to make a tiny PR that will return a subset of the loaded data, just so IGV can proceed without erroring out.

jrobinso:
@romanzenka I've never seen Chrome do this.  Are you sure its not your server?  It is an issue in many cases, for example a 300 GB bam file.

romanzenka:
I am using the built-in python server on a tiny VCF file. I guess it could be the server giving up, good point, will do more reading about it. It serves ranges just fine but once the ranges get small, it gives up.

jrobinso:
I think this is resolved, re-open if not.

PR #1050, opened 2020-01-24
Instead of erroring, provide slice of loaded data

romanzenka:
I am getting ArrayBuffer from the response which has a convenient .slice method that is useful for this sort of situation.

Fixing issue #1049

### DIFF
--- a/js/igvxhr.js
+++ b/js/igvxhr.js
@@ -200,7 +200,8 @@ const igvxhr = {
                         if (xhr.status === 0 || (xhr.status >= 200 && xhr.status <= 300)) {
                             if (range && xhr.status !== 206 && range.start !== 0) {
                                 // For small files a range starting at 0 can return the whole file => 200
-                                handleError("ERROR: range-byte header was ignored for url: " + url);
+                                // Provide just the slice we asked for, throw out the rest quietly
+                                resolve(xhr.response.slice(range.start, range.start + range.size));
                             } else {
                                 resolve(xhr.response);
                             }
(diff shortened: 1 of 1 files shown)


jrobinso:
This is nice, but how do we handle the not-uncommon case of someone trying to use a simple server with no range support to serve large files?    We want to discourage that.  Perhaps we could add a check on the range requested, if its > some threshold we error.   Is that reasonable?

romanzenka:
I think that erroring out because user did something that is sub-optimal is pretty harsh. 

But then I do not deal with the users reporting low performance...

What if we limit the file to ... I do not know, 100 kilobytes? If it is more than that, you get an error, if less than that, we let it slide?

romanzenka:
I added the check for the amount of data loaded being small <100kb

jrobinso:
I was thinking more like 100 mb, or at least 1 mb.    Thanks for the PR this is an improvement.

jrobinso:
OTOH if the response data is already in memory and the browser hasn't frozen I guess its not a problem.    I still don't understand why you are seeing this, we have unit tests that request a single byte.  I have seen the issue in some organizations that use proxies that strip the range header from the outgoing request.

jrobinso:
@romanzenka So I pushed a change to warn user if file is > 100kb, and never error out.   If performance tanks maybe they'll remember the warning and investigate further.

romanzenka:
@jrobinso Thank you! I am still investigating to understand exactly why this is happening. I have seen it in two instances - once with a simple python server, and once with built-in IntelliJ server.

jrobinso:
I know the built-in intellij server does not support the range header.
 Flask can support it,  I think it does by default know,  I don't know
about other python servers.

PR #1073, opened 2020-03-04
Add for support for a GWAS track based on data in a bed file.

benralexander:
To distinguish this new GWAS track code from older versions of the GWAS track, I added a new format designation called bedgwas.  There was some older code in gwasTrack.js which I believe was relevant only to an API-based version of the GWAS track that Jim had created for the Broad T2D portal project 3 or 4 years ago.  I'm guessing that this old code is not used by anyone anymore, but I left it in place just in case ( though that old code won't be used if you set the format == 'bedgwas' ).

In addition to plotting by -log pValues, the new code is also capable of handling posterior probabilities, which all have values between 0 and 1, and for which higher values indicate a stronger association (and are thus higher up in the graphical display).  To support this mode the code recognizes two new configuration values,
                               posteriorProbability: true,
                               ppa: 'score',
The first merely indicates that we are expecting posteriors, and the second value names the field in which to find the floating point value.

The reason to make this change was that the previous dependence on a range-based API call was always a performance bottleneck, and not nearly as good as depending on indexed BED files.  Therefore the new code expects the data to be provided in standard BED format.  The only convention we had to employ was that we needed to name the reference and alternate allele for each base, so we put those in the name column, separated by underscores.  An example of desired format of the bed file that drives the track might look like this:

chr1    533179  533180  A_G       0.62
chr1    544584  544585  C_T       0.28
chr1    565470  565471  G_A       0.62

That's it.  I avoided changing anything else in the file (though I did change the track to recognize the igv autoscale option).


jrobinso:
Thanks Ben.  I don't understand the need for a new format,  isn't enough to set "type=annotation / format=bed"?  The format only affects the parser.   In fact I don't see where the format specifier "bedgwas" is used.

jrobinso:
@benralexander the only other potential user of this track is the GTex portal.   Can you confirm with them if they use it or not,  if not old support for the T2D portal that is no longer used can be removed.

jrobinso:
@benralexander Could you include a small test file with the PR? You can put it under "test/data/gwas".   Thanks.
Jim

benralexander:
> Thanks Ben. I don't understand the need for a new format, isn't enough to set "type=annotation / format=bed"? The format only affects the parser. In fact I don't see where the format specifier "bedgwas" is used.

The "bedgwas" format is only referenced inside gwasTrack.js, and only then to differentiate the processing in this commit from the old gwas track code.  The code previously contained a format switch for "gtexgwas", but anything else was treated with the T2DVariantSource source.   If the old-style processing undertaken by T2DVariantSource is no longer used then we could get rid of the bedgwas format.

> @benralexander the only other potential user of this track is the GTex portal. Can you confirm with them if they use it or not, if not old support for the T2D portal that is no longer used can be removed.

I've reached out to the GTEx team with the above question.  They are definitely using IGV for eQTL plots, and therefore I would conclude that they are likely using the gwasTrack module.  Unfortunately, they tell me that the person who did the integration work (Duyen Nguyen) is out on vacation for next week, so I can't be entirely sure about their implementation until her return.  I will follow up at that time.

> @benralexander Could you include a small test file with the PR? You can put it under "test/data/gwas". Thanks.

Will do.

jrobinso:
@benralexander OK, thanks for the clarification, but I think you can just test for format "bed" there, you shouldn't need to introduce a new format.  I'd like to minimize format proliferation, there are around 50 in IGV desktop it becomes a maintenance burden.   Even though in this case it is not really a new format you can never be sure about that.

jrobinso:
@benralexander Can you give me a test html file that uses the new bed format for GWAS, preferably a whole genome view.   Thanks.

jrobinso:
@benralexander I ask because as far as I can tell the GWASTrack is not used, anywhere, perhaps at one time it was used in the Gtex portal.    So I don't know what to do this this, but I definitely could use more test data.    When most people think "gwas track" they expect a manhattan plot.  Perhaps that's what this is supposed to do.

jrobinso:
@benralexander OK disregard the above.  I've started over, basically, on this to support Manhattan style plots as IGV desktop does.  Please pull from the "gwas" track and test with your files.   I have a few more features to add to this before merging, but this will replace the current GWASTrack and related classes.   It will work with your bed format (type="gwas", format="bed") as well as other formats.

benralexander:
Hi Jim.  Okay, that's even better, of course.  Some recent developments on my end:

I did finally get a hold of Duyen Nguyen of the GTEx team, and she explained that they they call IGV.js with the configuration settings:
type: 'GTEX',
sourceType: 'gtex-ws'
Looking at the code I see that this approach generates a very similar plot to the association plots that we need for the T2D portal, but they are using distinct set of classes in the GTEX folder.  This is maybe another good reason to revisit this functionality, so that a common base of code can control at least the display and pop up methods for both groups.

Regarding your request for a file that could drive a whole genome Manhattan plot, I did come up with a gzipped bed file (along with an associated tbi) that should do the trick. The data file is about 1.1MB, and the index file is about 300KB.  I'm happy to upload them if you like.

On the topic of whole genome Manhattan plots, that is something that would definitely be useful for the project.  (Our current approach to Manhattan plots is based on strraight SVG, not HTML Canvas, and therefore it overloads pretty quickly if I put too many points into the plot.)  Note, however, that genome-wide plots are only one use case for us.  Most of the time people want to look at a much more restricted range of the genome, usually less than 2Mbases.  If they can zoom in and out from genome-wide to restricted range and back again then that would be really great.

So, bottom line, your plan to rework the handling of the association (and eQTLs, I'm assuming) sounds great, and we will find that functionality very valuable.  I hope that the ability to support the posterior probability plots (in which the vertical axis is not negative log scaled) is also retained, since users like that capability as well.  Thanks for your support and all of this, and please let me know if there's anything that I can do to help. -Ben

jrobinso:
@benralexander merging the gtex & gwas folders is a good idea, but not in the cards for me in the near future.  Pull from the "gwas" branch when you have a chance and test it with our data,  I think it will do what you want.

WRT whole genome manhattan plots,  these are only possible of course if you load all the data (whole genome after all), thus its not going to work with an indexed file.

jrobinso:
Oh is some example html in the "dev" folder,  gwas-bed.html and gwas.html.

jrobinso:
@benralexander GWAS support has been merged, and documented in the wiki.  It will be released in the next week or two in v2.5.0.

Issue #1102, opened 2020-05-05
Support for adding custom tracks dynamically


cklamann:
Would it be possible to dynamically create custom tracks, for instance by adding items to `TrackFactory` via the `Browser` instance? This could allow some users to 'plug in' their own tracks without having to maintain a fork. We have put together a simple implementation in our fork. The change would involve adding getters/setters to `TrackFactory` and creating something like an `addTrackToFactory` instance method on the Browser object. Thanks for your time!

jrobinso:
That's an excellent idea.   Do you want to send a PR?

I have to warn you that I am swamped for the next month writing a grant renewal for IGV / igv.js and might not be able to review and merge it until that's done.   But I do like the idea.

cklamann:
Great, I can submit a PR probably by the end of the week. And no problem about the turnaround.

PR #1105, opened 2020-05-11
allow browser instance to add tracks to TrackFactory

cklamann:
refs issue #1102


jrobinso:
@cklamann I've not forgotten this, heads down on the IGV renewal grant.  Will get to it ASAP

PR #1175, opened 2020-09-11
Skip empty lines in data file

dedaimia:
Little background on this PR:
We have users who habitually insert blank lines into files (mostly .bed but other file types as well), because they like how it looks when reviewing in a text editor.

Currently, igv.js parsing code will stop when it hits a blank line.  Features before the blank line are loaded, the rest of the file is mysteriously (from the user's perspective) missing.       No error message.

Initial reaction from our developers was to insist they remove empty lines, but it was pointed out to us that all other browsers they use (UCSC browser, IGV thick client, Alamut) handle them fine.  Also users are very attached to them.  And it does seem erroneous to load only part of the file without giving an error message.

Proposed code change to fix this:  change StringDataWrapper to skip empty lines in a data file, instead of returning undefined/empty string.    The old behavior (undefined/empty) would cause most featureParsers to stop reading the file when while (line = nextLine()) evaluates as false on the empty line.


jrobinso:
I agree this is a problem that needs addressed, but I'm not sure I want to fix it in this way.   There are situations where the # of lines are counted, and this would throw that off.    Feature parsers should continue past blank lines, or lines that start with comment delimiters (e.g. #),  so that is where the fix should be.   I will look at what's going on there.

dedaimia:
I'm not entirely sold on this fix either, it was just a thought.  My first impulse was to fix in the feature parser as well, but it looked like there were a number of feature parsers with the same "while (line = dataWrapper.nextLine())" check that is going to stop looping as soon as DataWrapper returns undefined or an empty string.  Pretty sure the intent there was to stop at end of file, but empty line stops it too.    So the DataWrapper change was just a quick way to get them all.   If fixed in feature parser it's going to hit several parser files and need to be fixed twice in each (in parseHeader and parseFeatures).

What would be REALLY nice (but much larger) is if the feature source could give us some kind of summary information on the file, like:
headerLines: 2
blankLines: 4
validFeatures: 1000
invalidFeatures: 10 (this may have changed since last time I checked, but we were noticing that some invalid features, like for a chromosome that does not exist due to user typo, seem to "disappear" as well)

jrobinso:
Yes I'm working on a fix.   In javascript the empty line is "falsey",  I think, but there might be other issues.    RE "invalid feature", I'm not sure what you mean by disappear.   A malformed feature line will be skipped.   A misnamed chromosome will be returned but you'll never see it obviously.

dedaimia:
Yeah, null, undefined, empty string are all falsey in javascript.   
Looks like the current dataWrapper can give back either undefined or and empty string for visually empty line, depending on whether line is just \n or has some other whitespace characters first like \t\t\n.   
Should be consistently returning undefined at end of file, but I didn't actually test that, just basing on code examination.

jrobinso:
This should be fixed now with commit ae33ba2f

dedaimia:
Sweet!  Thank you.  That change will work for us (and much more along the lines of what I originally thought for a fix too).  What about all the other places that do a "falsy" check on dataParser.nextLine() though?  I see it in all of these:
AEDParser.js
gcnvParser.js
genBank.js
gwasParser.js
rnaStruct.js
segParser.js
vcfParser.js

Not sure that's all of them, could be missing if someone used a different name for their dataWrapper instance.   May not matter in some of these cases (empty lines perhaps not possible), but I know I've seen people pull the empty lines trick on vcfs.

jrobinso:
Yes they all need fixed.   We've not encountered many empty lines in file actually, obviously this just came to light after many years.

jrobinso:
I changed this in places that were really testing for empty lines.     I am not sure a blank line in a VCF is legal and you might have problems with that in other tools, but it should load here.

dedaimia:
Thank you!  We weren't expecting a decision to my PR anywhere near this quickly, let alone a fix, your responsiveness is impressive.  I think it's pretty questionable whether empty lines are valid in any of these files, and they most certainly do cause issues in other tools (e.g. tabix won't index a file with empty lines), but at least igv.js is now consistent with all the other visualization tools, which seem to accept them.

PR #1180, opened 2020-09-19
Unit test conversion

jrobinso:


jrobinso:
rewrite unit tests for mocha / chai

Issue #773, opened 2018-11-02
Display of long text string in Annotation track pop up


keithforest:
I would like the user to be able to view a long text string from the 'Note' field of a gff3 file in the popup that occurs when hovering over a transcript.  Right now, the popup cuts the note field after ~30 characters.  Is there a way to show the entire text string?

jrobinso:
Not currently,  I tagged it for the next release.     Put any thoughts on how to do that here.

jrobinso:
@turner @helgathorv we need a solution for this.

turner:
Closed with commit b1e948b

Issue #846, opened 2019-01-24
Enable Display of Verbose Annotation Names (e.g. via Hover)


dicara:
We want to display user generated annotation labels that are often quite verbose. Even if a user clicks on the annotation, the name is abbreviated in the pop-up. As opposed to left/right labeling (as was suggested in https://github.com/igvteam/igv.js/issues/385), I was wondering if hover could be implemented. In other words, if a label doesn't fit under the annotation track, add as much of the label as will fit and suffix the label with an ellipsis to indicate it has been abbreviated. When a user hovers over the track, then the entire annotation label would appear.

turner:
Closed with commit b1e948b

PR #1203, opened 2020-10-22
Track popover edits - tooltips & selectable text

turner:
Track popover edits - tooltips & selectable text


jrobinso:
Fixes issues #773 and #846

PR #1234, opened 2020-11-25
Set max size to start of VCF data to ensure read all headers

RoadRoller:
I face with issue that fixed maxSize 65000 maybe less than actual VCF headers size, this pull request fix it by evaluating maxSize from index data, so it will be set to start of data section and all headers will be read. Looks like the only problem was with tribble index, for tbi it was already implemented.

Some point that can be improved:
- I don't know is index data guarantee order of chrs and blocks, if it ordered, code can be simplified by just selecting first chr and first block
- Maybe I should add test for this case, it's my first touch of this project, if test is needed, please point me to similar test example


jrobinso:
Thanks for the PR, this looks like a good improvement.   I will not be able to review it until next week.  

A unit test would be an excellent idea.  Unfortunately there is only the most minimal test for "tribble", basically that a tribble index can be read without error.  It is in test/testTribble.js,  you could add a test there.   There are more extensive tests for tabix in testTabix.js that could be helpful for a model.  Tests are designed to run in node, without a browser, with mock objects used to simulate the browser when needed.   In this case that shouldn't be neccessary.   If using a local data file for testing note the use of ```require.resolve```, as in. 
```
url: require.resolve("./data/tabix/csi-test.vcf.gz"),
```

jrobinso:
@RoadRoller I merged this after some modifications, the most important being that to guarantee you get the entire header block.max must be used, rather than block.min.   For files with small headers the first block.min == 0.   

There is a unit test for a file with a small header, but not a large one that motivated the PR to begin with.   If you could provide a test file I will add a test.

RoadRoller:
@jrobinso, thank you!
Here the file that can be used for unit test:
[test.vcf.gz](https://github.com/igvteam/igv.js/files/5633538/test.vcf.gz)

js/feature/featureFileReader.js
```diff
@@ -108,14 +109,17 @@ class FeatureFileReader {
...
-                const dataStart = index.firstAlignmentBlock ? index.firstAlignmentBlock : 0;
-
                 let dataWrapper;
                 if (index.tabix) {
                     dataWrapper = new TabixBufferedLineReader(this.config);
                 } else {
+                   // Tribble
+                   const maxSize = Object.values(index.chrIndex)
+                        .flatMap(chr => chr.blocks)
+                        .map(block => block.min)
```
RoadRoller, line 119: @jrobinso as I understand from you comment, we need to use `block.max` here?

Issue #1238, opened 2020-12-03
Feature request: allow listeners for track order changes


bw2:
I'd like to be able to listen for track order changes, and get the new track order from the event (maybe as an array of track view objects?).  This would allow me to save/restore this from cookies, etc.

PR #1249, opened 2021-01-03
Track order changed event

bw2:
Implements #1238


jrobinso:
There's a possible issue lingering here in that the "name" property of track is not guaranteed to be unique.  A more robust and general return value might be just and ordered array of track names.    Something like

```
trackViews.map(function (tv) {return tv.track.name}
```

PR #1263, opened 2021-02-11
Bw2 update track config method

jrobinso:


jrobinso:
@bw2 Oops this broke painting of Y axis for wig type tracks.

PR #1267, opened 2021-02-18
wrap event value in array, filter out empty or null track names

bw2:


jrobinso:
Looks like we fixed the same problems with the original version?

bw2:
yeah, I didn't catch that you fixed this already.

PR #1281, opened 2021-03-05
Allow HTML in the popover for variant custom reader

amanpatel:
We needed a way to create a custom variant reader and have a button to view more information in the popup dialog. This enables that by doing something like:

` fields.push({ html: '<button>Some button for this variant</button>'});`

as part of the `Variant.prototype.popupData` function in the custom reader.

Also, this "feature" would also get rid of the hardcoded checking for "<hr>". Although that does break compatibility for anyone who relied on this (no hurt in keep it there).


jrobinso:
Thansk for the PR, looks good.  Sorry for the delay.

PR #1327, opened 2021-05-12
Fix bug: Empty text features list causes error

lukasjelonek:
I use igv js for displaying features in annotation-tracks via the `feature`-option. When a track-feature-list is empty an error dialog is displayed.

Example code:
```js
const config = {
        reference: {
          id: "my genome",
          fastaURL: this.fastaUrl,
          indexed: false,
          tracks: [
            {
               name: "empty",
               type: "annotation",
               features: [],
            }
          ],
          wholeGenomeView: false
        },
      };
```

Error:
image

The error appears in the `FeatureCache` constructor, but is caused by not returning a list from `fixFeatures`. This patch fixes the `fixFeatures` method. It might also be an option to check if the `featurelist` object is `undefined` in the `FeatureCache`-constructor.

### DIFF
--- a/js/feature/textFeatureSource.js
+++ b/js/feature/textFeatureSource.js
@@ -313,7 +313,7 @@ function packFeatures(features, maxRows) {
  */
 function fixFeatures(features, genome) {
 
-    if (!features || features.length === 0) return;
+    if (!features || features.length === 0) return [];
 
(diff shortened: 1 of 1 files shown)


jrobinso:
Looks good, thanks!

PR #1379, opened 2021-09-17
rulerViewport: check tooltip defined in mouseMove/setTimeout

f-krull:
Problem: After igv.js is closed, a setTimeout might trigger, which is then tying to read an undefined `$tooltip` property.

Selection_381

This PR checks if the `$tooltip` property is defined.


jrobinso:
thanks

PR #1385, opened 2021-10-01
Add an option to disable continuous update of a track's content while panning

GMignerot:
Currently, IGV.js updates the content of the tracks continuously while the user is panning.
This may trigger multiple simultaneous, redundant feature reads and redraws while the user is panning on the browser, as each `mousemove` event reloads the whole visibility window, around the same width on each side, and redraws the whole thing.

While this makes perfect sense for local files or small amounts of data, when IGV is integrated in a web app and dynamically consumes relatively large amounts of data from a remote server, this behavior rapidly creates multiple heavy requests while the user is dragging the browser, that are a huge burden on network and performance, both for the client and the server.

This pull request adds a property in the **track** configuration (to be able to disable this only for that kind of critical tracks) that allows configuring this behavior, called `panContinuousUpdate`
If `panContinuousUpdate` is set to `true` or not defined in the track config, it keeps the usual behavior, updating the track content as the user is dragging.
If it is set to `false`, it disables the content update and redraw while the user is still dragging and only updates the data when the user releases the mouse drag, and thus makes only one feature read for the whole panning, greatly reducing redundant requests.

This config option is set to `true` by default, to keep the existing behavior when not defined.

### DIFF
--- a/js/trackBase.js
+++ b/js/trackBase.js
@@ -89,6 +89,8 @@ class TrackBase {
 
         this.visibilityWindow = config.visibilityWindow
 
+        this.panContinuousUpdate = (config.panContinuousUpdate === undefined) ? true : config.panContinuousUpdate;
+
         this.trackView = undefined
     }
 
--- a/js/trackView.js
+++ b/js/trackView.js
@@ -373,6 +373,7 @@ class TrackView {
      * Update viewports to reflect current genomic state, possibly loading additional data.
      */
     async updateViews(force) {
+        const isDragging = this.browser.dragObject;
 
         if (!(this.browser && this.browser.referenceFrameList)) return;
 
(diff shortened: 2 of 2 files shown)


jrobinso:
Actually what you describe is not correct.   igv.js draws to an offscreen canvas and that is panned with mouse moves.   Data is loaded only occasionally,  and only as needed to draw a new offscreen canvas.

GMignerot:
Indeed, I did not word it properly.

What I’m saying is that every time the user is panning beyond the currently loaded range, it triggers a data read. With local files and small amounts of data, it’s not an issue as the process is almost instantaneous.

But when dynamically loading relatively large amounts of data from a remote server, the request, data processing and redraw take time, and meanwhile the user can still move further beyond the loaded window, triggering multiple new reads while the others are still pending — each read being for a full window, only shifted by a few bp.

So this adds an option to prevent having that much redundant requests simultaneously, while still retaining the existing behavior for less critical tracks.

jrobinso:
OK this looks reasonable I'll try to get to it this week, but I am traveling so it might be delayed.

jrobinso:
@GMignerot I just took a quick look, mostly o.k. but there seems to be some unrelated reorgnization of code related to computing the max row count.   Was this done to handle some situation for multi-locus views?   In multi-locus view the user can be panning one view, but the other views will be affected on data load from any view as the number of rows (overlapping variants) could change.   I don't necessarily see a problem with the changes I just want to understand what motivated them.

jrobinso:
An update on this -- looking more deeply into this, as well as testing the PR,  i am leaning towards the view that this should be the default behavior, or perhaps only behavior.  It was never really my intention that round-trip calls to the server for data would occur while a user was actually panning.    However if data has already been loaded, for example from a non-indexed file, then repainting should occur,  so its not as simple as just taking the PR and defaulting to "true".    At one time, in a previous release long ago, we would freeze panning while any reloading was in progress.   I think that was removed, it has its own issues, and that's how we ended up with the current state.

GMignerot:
Well, I'm happy this could help !
This works perfectly, so I think we can close this
Thank you so much for all your hard work !

Issue #756, opened 2018-10-17
which function may I call to trigger redraw?


paul-shannon:
In my shinyApp, I sometimes rearrange widgets, sometimes giving igv a large portion or all of the screen, sometimes letting a DataTable dominate the display, sometimes hiding one or the other.  

igv does not seem to fully recognize this resizing, which I accomplish by changing the css class of its parent element.  When I manually resize my entire web browser, igv and  DataTables both arrange themselves nice.   Apparently manual resize triggers a cascade of redraw events which work their way down to the appropriate functions in igv and DataTable.

Is there an igv redraw function I can call directly?

jrobinso:
Resizing dynamically by changing CSS classes is not going to trigger any
events igv.js can receive.   Try calling igv.visibilityChange(),  or
browser.resize();

paul-shannon:
Solved, using your suggestion.   Thanks, Jim.

Hamza-Yousuf:
i solved this issue by this browser.resize();

PR #1429, opened 2021-11-22
add clickToHighlight to gCNV track

### LINKED ISSUE #1411: Highlight samples in gcnv bed on click instead of specifying in file
It would be great if when loading bed files of copy ratios from gcnv if the highlighting on samples could be done 'on click' instead of having to specify these either in loading config or written into the first line of the bed file.

At the moment I know it is possible to add the following:

`track  type=gcnv height=500 highlight=sample-X12345;red\n`

image

But when wanting to load these bed files into any instance of igv.js/igv-webapp this isn't convenient as the sample(s) to highlight have to be determined when creating the bed file.

Also, would it be possible to get better docs on the use of gcnv, I only knew about this from my internship at the Broad and using tgg-viewer, and it would be really helpful for sharing this wider as it is very useful for cnv visualisation.

Thanks!

jethror1:
Adds clickToHighlight option to gCNV track for setting sample line colour on click.

Accepts either a single fixed colour or a keyword `any` that sets a random colour for each line.

Line colour is unset on clicking the same line again.

Closes #1411 

Example of `clickToHighlight=red`:

gcnv_igv_red

Example of `clickToHighlight=any`:

gcnv_igv_random


jrobinso:
This is a rather extreme method to force a repaint
```
igv.visibilityChange()  // prompt redraw to change colour of clicked sample
```

Try this instead

```
this.trackView.repaintViews();
```

Where "this" is the track object itself

jethror1:
Have changed it, I wasn't sure on the best way to prompt redraw and used your suggestion from here: https://github.com/igvteam/igv.js/issues/756#issuecomment-430672076

jrobinso:
Yes I should document this,  apologies for that.   resize() and visibilityChange() will work but also potentially result in repainting all tracks, not just the one affected, and also potentially a data reload, not just a repaint.

PR #1451, opened 2022-01-26
Further abstract variant module

haoyuanli:
It's a simple pull request that further abstracts out the `Variant` class, this
- makes it easier for future additions of classes or third-party implementations that implement `Variants`.

Based on the pattern here, I made a child class of `Variants` called `VCFVariants`, which is used for the function `createVCFVariant`.

The code changes here are merely refactoring, and should not change existing behaviours.

### DIFF
--- a/js/variant/variant.js
+++ b/js/variant/variant.js
@@ -30,22 +30,21 @@ import {StringUtils} from "../../node_modules/igv-utils/src/index.js"
 const knownAltBases = new Set(["A", "C", "T", "G"].map(c => c.charCodeAt(0)))
 
 function createVCFVariant(tokens) {
-    return new Variant(tokens)
+    return new VCFVariant(tokens)
 }
 
 
 class Variant {
 
-    constructor(tokens) {
-        this.chr = tokens[0] // TODO -- use genome aliases
-        this.pos = parseInt(tokens[1])
-        this.names = tokens[2]    // id in VCF
-        this.referenceBases = tokens[3]
-        this.alternateBases = tokens[4]
-        this.quality = tokens[5]
-        this.filter = tokens[6]
-        this.info = getInfoObject(tokens[7])
-        this.init()
+    constructor() {
+        this.chr = null
+        this.pos = null
+        this.names = null
+        this.referenceBases = null
+        this.alternateBases = null
+        this.quality = null
+        this.filter = null
+        this.info = {}
     }
 
     init() {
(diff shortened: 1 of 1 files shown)


jrobinso:
@haoyuanli thanks for the pull request.   Do you have a specific use case for this?  If so I would like to discuss solutions.   I think the approach in this PR does not really add any flexibility,  and could lead to confusion.   The Variant class is not in fact designed to be abstract, the concrete implementation is required for VariantTrack to work (i.e. all or most of the fields that are set to "null" in the constructor must be supplied.     So a Variant must be created by some means, the function createVCFVariant does it from a VCF file.   In the past there was also a createGA4GHVariant that created the same object from the now defunct GA4GH webservice protocol.   

One oddity your PR brings to light, which should be changed, is initializing the Variant object in the constructor from the fields obtained from parsing a VCF file.   I will look into moving those out, but in the end those fields have to be set.

As javascript is not a typed language there is nothing preventing another implementation of "Variant" to be substituted, but again the fields as they exist must be implemented in the other implementation.   I think it more like to want to create alternative factory functions (e.g. createVCFVariant) to create that object.

haoyuanli:
Hi @jrobinso 

Thanks for the detailed response!

The reason I made the changes was because I was making some custom changes on top of the now-deprecated `createGAVariant()` method that exists within IGV.js (that we still use to some extent).

image

Since the method calls to initialize a `Variants` class this way without passing a `tokens` value into the constructor (which it doesn't have), it would introduce an exception.

Therefore, I wanted to move the `tokens` out of the `Variants` class, as the `tokens` is specific to the `VCFVariant`

I hope that makes sense! I keep the changes in a separate repo, so they don't really need to be merged, but I thought someone might run into similar issue in future (e.g., initializing the `Variants` class without necessarily having the `tokens`.

Hope that at least clears up the initial motivation of the changes!

Best

jrobinso:
@haoyuanli Thanks for the clarification, yes I see that the way its constructed now would be a problem.    I will refactor this a bit, probably back to what it was when GA4GH was supported.   Since the issue is how a Variant is constructed, not specialization to the Variant class per se,  I'd rather keep this generalization at the construction stage.

jrobinso:
@haoyuanli I have moved the tokens based initialization to the createVCFVariant function, leaving a blank constructor.   This is how it must have been when GA4GH was supported.  Let me know if this solves your issue.

haoyuanli:
Thanks @jrobinso , that would work well. I will close this PR!

image

One minor thing, is that I guess we don't really need to pass the `tokens` into the `Variants` class as it no longer consumes one.

Issue #864, opened 2019-02-21
support bgzip reference file


odoublewen:
Feature request:  
It would be nice if igv.js supported bgzipped fasta files for the "reference" object.  (I assume it is not supported, as it is not mentioned [here](https://github.com/igvteam/igv.js/wiki/File-formats) and also I've tried it without success.)

Background: `samtools faidx` will happily create an index on block gzipped fasta files.  It actually creates two:  the standard `.fai` one, and another `.gzi`, which [marginally documented](https://github.com/samtools/htslib/issues/473).

You have bgzip + tabix working, so seems like this might be feasible too.

jrobinso:
Yes that's supported in IGV desktop.  Its moderately complex,  I don't know how many people use it,  and you are the first to request it.   I would consider a pull request for this, but it won't get prioritized anytime soon.  I agree it would be nice.

cmdcolin:
just for ref https://github.com/GMOD/indexedfasta-js supports bgzip fasta :)

jrobinso:
Thanks @cmdcolin I might just use that.  You've got pretty much everything there,  I would contribute some modules if there's anything you are missing.  VCF maybe?

odoublewen:
@jrobinso Thanks - I agree it needn't be high priority, just a nice to have.  

@cmdcolin Thanks for the reference. I'd used jbrowse many years ago, it's looking good these days.

cmdcolin:
Grazi :) we started these separate npm modules for the data types as part of a big refactoring. We have a vcf module already but thanks. I liked the .hic adapter igv has though!

jrobinso:
@cmdcolin You're just a little ahead of me,  I have been planning to do the same thing with igv.js this spring.   I'm on the fence on whether to use Node or ES6 type imports.   I would like to use ES6 if Node had better support for it.     RE hic adapter, are you referring to hic-straw or something else?

cmdcolin:
Yep, the hic-straw module https://github.com/igvteam/hic-straw

cmdcolin:
I think it's probably worth just use the ES6 type import/exports, but it's probably fine to do require too. If you have some level of babel it tends to just get smoothed over some way or the other.

Some people are opposed to the "export default" syntax in the ES6 modules for various reasons but you can use ES6 modules without export default too.

jrobinso:
[...]

PR #1452, opened 2022-01-27
Implemented support for reference genomes compressed with the BGZIP a…

PSB-mibel:
…lgorithm. Implementation was derived from the GMOD-bgzf package, and reimplemented to make use of the igvxhr interface. To make use of the compression, simply add the compressedIndexURL parameter to the reference object that is provided to igv.js. Important note: the current implementation only works when the reference genome is indexed! Compressed by non-indexed genomes are currently not supported.

As indicated in issue #864 I've mostly made use of the work from the @GMOD team (https://github.com/GMOD/). 
As both projects (IGV.js and GMOD) are released under the MIT license, I think licensing wise this copying/deriving of code should be fine. 

I've tried following the coding style from the original file (e.g. no semicolons etc.), and in the end I didn't need to use any additional packages, so building the project using Rollup should work fine. 

If there's any issues related to this patch, feel free to contact me: mibel@psb.ugent.be


jrobinso:
@PSB-mibel Thanks this will be appreciated by many.   I will merge it as soon as I'm able.

jrobinso:
Apologies for the delay on this, I had some unexpected interruptions and then a lot of IGV desktop issues to work through.  This is now my priority for the rest of the week.

jrobinso:
@PSB-mibel this is merged.  And BTW this was a very nice PR, existing code was touched minimally and it was very easy to review.   Not all are like that, thanks.   I have added a unit test which I will push shortly.

PSB-mibel:
Thanks for the compliment. I hope the changes pass the unit test.
I've already deployed my build of IGV.js with support for the bgzip ref genome, and everything does seem to be working fine up until now, so I don't expect the unit test to fail. However, if it does, I'll be happy to try and help debug it.

jrobinso:
Its working, already merged to master.   This will be in the next release.

PR #1460, opened 2022-02-12
Multi Track Selection and Functionality

turner:
- Track Height
- Track Color
- Track Merge
- Track Removal
- Auto Group Scale


julianstanley:
I use two wig files that represent 5'-end and 3'-end RNA sequencing counts. In regular IGV, I overlay them in one track like this:

image

I was trying to figure out how to do this in igv.js when I saw this issue. When you're talking about "track merge" here, is that what you're referring to? Or should I make a separate feature request for that? (Or does it already exist?)

PR #1461, opened 2022-02-14
bug fix - IGV js hangs and crashes browser if something is "off" in the url

eyal-elkevity:
bug fix - when loading the igv, if you don't provide an explicit `queryParametersSupported: false` in the options object, the igv iterates over current query parameters looking for parameters of certain names (like 'file') and appends them to the config. and if, by any chance, you have corruption in the url (like we had an accidental `&&` in the middle), the igv loading hangs the browser (goes into infinite loop). this simple fix solves this (if a query param is somehow invalid, move on to the next)

### DIFF
--- a/js/igv-create.js
+++ b/js/igv-create.js
@@ -235,6 +235,8 @@ function extractQuery(config) {
                     config[key] = value
                 }
                 i = j + 1
+            } else {
+                i++;
             }
         }
     }


jrobinso:
Thanks for the fix.  There might be another bug here as well,  queryParametersSupported should default to false.   Will look into that.

eyal-elkevity:
Hi Jim and thanks for the merge! Yes, I feel the same about the
queryParametersSupported config, I was surprised to see it defaults to
true, because if by any chance I'd had a query param with a matching name I
would get a weird config and would not understand why... feels like it
should be an explicit configuration. Actually my fix was to set this to
false, but after finding the bug in the igv code I felt the need to create
this patch. Thanks again :)

jrobinso:
Yes they were both bugs,  it defaults to "false" now.  Thanks again for the fix.

Issue #1498, opened 2022-04-18
Is it possible to make Insertion look similar to the desktop version?


minmarnoo:
In Desktop version, the INS is more highlighted and also indicates the number of bases inserted (in this case, 6).

![image](https://user-images.githubusercontent.com/86052341/163741885-3a6395ff-5512-44cc-a4e5-1d446b1e4ecf.png)

![image](https://user-images.githubusercontent.com/86052341/163744451-dade7992-368f-48df-bf02-4788e7cd3332.png)

jrobinso:
All things are possible of course, not all things are implemented.   Leave this open as a request.

PR #1526, opened 2022-06-16
allow the display of insertion/deletion base counts

davidlougheed:
addresses #1498 

Adds 4 new config options for alignment tracks:

* `insertionTextColor` (default: white)
* `showInsertionText` (default: false)
* `deletionTextColor` (default: black)
* `showDeletionText`: (default: false)

if both insertion and deletion text are shown via the above options, it may look something like this (example with CCS reads + HTT expansion; ignore the read colours):
Screen Shot 2022-06-16 at 6 09 41 PM

there is also a bit of logic to decide when to stop drawing these labels when the user is zoomed out enough.

this also changes bases to be drawn on top of insertions/deletions to match the Java version. this can be reverted fairly easily if desired. gaps are now also drawn underneath insertions to display more intuitively.

If this PR is merged, I agree to transfer the copyright of these changes to the copyright holders of the `igv.js` repository.


jrobinso:
Nice, thanks!

Issue #1366, opened 2021-08-23
Support for encoded data of binary files


gaitat:
I would like to ask if there is support for base64 encoded data (i.e. data:application/octed-stream) of binary files. For example I have a case of bigbed and bigwig files that I need to encode before passing them to igv and they dont seem to work.
On the other hand I also tested base64 data encoded wig files and they worked fine.
For example this sequence works:
`const bwSource = new BWSource({url: require.resolve("./data/bb/myBigBed2.bb")});`
On the contrary this does not work:
`const bwSource = new BWSource({url: 'data:application/octet-stream;base64,6/KJhwQAAAADBAAAAAAAADMEAAAAAAAACBMB ...});` 

Thank you

jrobinso:
data URIs (including gzipped data) should work for fasta, bam, and plain text formats (e.g. wig and bed),  but will not work for cram or "bb" formats (bigwig and bigbed).  If you need to do that I suggest first converting to bed or wig,   then gzip (optionall) and create the data URI.   For some examples see the files in. "dev/datauri".

gaitat:
Is it a technical limitation?

jrobinso:
I'm not sure what you mean, its not implemented, of course it could be done.

jrobinso:
This should work now, but it has not been exhaustively tested.    See https://github.com/igvteam/igv.js/blob/master/dev/datauri/ucsc-interact3.html for an example

gaitat:
Thank you, will do some testing.

PR #1548, opened 2022-09-27
Remove cram data uri check

dennishendriksen:
In a [previous comment](https://github.com/igvteam/igv.js/issues/1366#issuecomment-904164965) it was mentioned that data URIs don't work for the cram format. Data URIs seem to work just fine for the cram format when removing the check that disables it, see the provided test case. You must provide an indexURI (which can also be a data URI), but I guess that is always the case for cram alignment tracks?

This makes sense since the [underlying code in igv-utils](https://github.com/igvteam/igv-utils/blob/v1.3.14/src/igvxhr.js#L103) handles data retrieval for both data URIs and other URIs.

We're currently using embedded bams in offline html files and would be really happy if we could switch to cram to reduce file size.


jrobinso:
OK, thanks for the PR. This has evolved over the years, originally the data URI was expanded in place and indexes were ignored, it gradually became more consistent with "regular" URIs

PR #1549, opened 2022-09-27
Local/global graphType  for mergedTrack and bugfix

arpanda:
Updated the merged track for the following.

1. Added the option for global graphType and If global setting is not present, local track configuration works.
2. Bug fix for supportsWholeGenome. supportsWholeGenome is now not property not function. 

Thank you
Arijit


jrobinso:
Looks good, thanks for the PR!

PR #1555, opened 2022-10-24
use nameField when parsing BED files

jvolkening:
The purpose of this patch is to honor the `nameField` configuration option when creating annotation tracks from BED files which use a `gffTags` name field.


jrobinso:
Thanks for the PR.  @malcook this might interest you.

malcook:
Thanks @jrobinso - should I expect to find this in nightly now or should I wait a day to try it out?

jrobinso:
Its merged, so if you are building from source it should be there, or
alternatively using the "snaphsot" build here:
https://igv.org/web/snapshot/

PR #1576, opened 2023-01-12
Multi locus label responsiveness

turner:


jrobinso:
There's a regression here, clicking the label should "goto" that locus.   That's what the current release does.

PR #1580, opened 2023-02-03
Implementation of cnvpytor track

arpanda:
track name: 
----------
cnvpytor

What it does: 
--------------
It allows one to view cnvpytor data

1. Can read a pytor file (CNVpytor produced file) and show
2. Read a vcf file and run CNVpytor method.

Example:
--------------
An example added at `examples/cnvpytor/cnvpytorTrack.html`

image


jrobinso:
Congratulations on the PR!   As most of the code is separate in the CNVPytor package I won't be doing a detailed review of it.   Here are some comments.

* The https import of the hdf5 library ```import h5wasm from "https://cdn.jsdelivr.net/npm/h5wasm@0.4.9/dist/esm/hdf5_hl.js"``` is causing the unit tests to fail.  As this is a very small file let's change that to a normal dependency,  list it in package.json,  and import it in the normal way by relative file path.  You can see lots of examples in many files, the import will look something like ```import h5wasm from "../../../node_modules/dist/esm/h5_hl.js"```.  It might not be exactly that but you get the idea.
* For consistency with the other code keep class names capital, so CNVPytorTrack or similar.  The file name can remain lower camel case.
* Files in the "examples" folder will be deployed automatically to a release server, so they should not import from the raw source code.  So things like this  ```import igv from "../../js/index.js"```  should be changed to ```import igv from "../../dist/igv.esm.js"```.     
* Re examples -- these are for users and developers of IGV to browse,   they should be well documented with notes and pertinent to your final release.   Some of these examples should maybe be moved under the "dev" folder.    Perhaps just 1 good example in "examples" would be enough, but if more are needed and relevant to your release that's o.k. too.
* RE documentation -- we will need some end user documentation for this track (or tracks?) for the wiki.   See the description of other tracks for examples.   It doesn't have to be elaborate, just cover the basics.  You can submit your markdown as a PR to the wiki repo https://github.com/igvteam/igv.js/wiki,   or just create it in your own wiki and send me a pointer.
* BTW the library you are using for HDF5 will read the entire file into user memory.    This should be o.k. as long as the files aren't too large, just noting it.

jrobinso:
Also, very important, users of igv-webapp (igv.org/app) will be loading these files without the benefit of a track "config".  The only thing igv.js is going to see from the config is the "url" property.   Most of your end users will be viewing files in this way (through igv.org/app), unless you build an application for them.   This means that everything has to be discoverable from the url to the pytor file.   Track type is determined from the file extension,   I think you've already handled that.  So "pytor" will be igv.js's signal that the track type is "cnvpytor".   Name is also derived from the url,  but if there a suitable name in your hdf5 file you could implement the track to use that.   If "points" is the preferred graphType your track should default to that.    The way to test that this will work for end users is to remove all properties from the config except "url" in your example.

```
            const pytorTrack = await browser.loadTrack({
                id: "pytor_track",
                type: "cnvpytor",
                name: "HepG2",
                graphType: "points",
                url: "https://storage.googleapis.com/cnvpytor_data/HepG2_WGS.pytor",

            })
```

arpanda:
Thanks for your comments. I have done the following to be consistent with the code and fix the build error. 
- `h5wasm ` is added to package.js and import is changed to `node_module` directory. Now importing using
```
import h5wasm from  '../../node_modules/h5wasm/dist/esm/hdf5_hl.js';
```
- Class name is updated to `CNVPytorTrack` 
- For examples, import changed to `../../dist/igv.esm.js`
- The `graphType: "points",` parameter is hard coded in CNVPytorTrack class. 
 
The auto build is failing due to `npm run test` command, but `npm run build` is working fine.

jrobinso:
OK thanks, I think the examples still need some work,  have you read them?  For example, this doesn't look up to date to me.   They need to be understandable by users unfamiliar with your track.

```
<ul>
    <li><a href="indexed_auto_update.html">indexed_auto_update.html</a> - Creates coverage automatically for the region in view from an indexed VCF file.</li>
    <li><a href="indexed_explicit_update.html">indexed_explicit_update.html</a>  - Creates coverage on demand at a user-specified bin size for the region in view.</li></li>
    <li><a href=not_indexed.html>not_indexed.html</a> - - Creates coverage over the entire genome (or chr22 in this case) on demand at a user-specified.</li></li>
    <li><a href=cnvpytorTrack.html>cnvpytorTrack.html</a> Reads a cnvpytor file and generate view</li></li>
    <li><a href=cnvpytorTrackVCF.html>cnvpytorTrackVCF.html</a> Read and process a VCF file for read depth and BAF information, and creates a view. User need to wait a bit as it process the input file on fly</li></li>
</ul>
```

The tests need to succeed before merging this,  I will look into the problem there.    The other thing we'll need is the user documentation.

jrobinso:
Yikes, problem -- the hdf5 import swells the size of the minimized igv.esm.js from 900kb to 4.2mb.  We can't accept a 4 fold increase in size for this track.   So I think we need to revert back to the https import

```
import h5wasm from "https://cdn.jsdelivr.net/npm/h5wasm@0.4.9/dist/esm/hdf5_hl.js";
```

and I will figure out a way to fix the unit tests.    Sorry, the hdf5_hl.js file itself is small but it must be pulling in some large dependency.

arpanda:
Sure. I have updated the hdf5 import and deleted old examples.

jrobinso:
Hi @arpanda , Could you change your merge target to bet the "arpanda-cnvpytorTrack_dev" branch in our repo, rather tha master?   You might have to close it an open a new one,  but just select that branch in the pulldown.  I've figured out how to fix the unit test problem caused by the https import of the HDF5 file.  However its a little complex and I'd like to do that in a branch.

arpanda:
Updated the brach from `master` to `arpanda-cnvpytorTrack_dev`

jrobinso:
@arpanda Is this ready to merge now?

jrobinso:
@arpanda I've merged, and built, but the examples aren't working for me.    The first one loads after about 2 minutes, but the track is blank.   The second give a time out warning and option to exit the page.   I chose "wait" and after another minute or so it appeared to load, but again no track just blank.  See screenshots.


Screen Shot 2023-02-03 at 2 34 25 PM

Screen Shot 2023-02-03 at 2 36 44 PM

jrobinso:
Also, there is no diffference in the 2 examples (cnvpytorTrack and cnvpytorTrackVCF),  they load exactly the same file with exactly the same options.

arpanda:
There was a scaling issue and I updated the branch.
- For the cnvpytorTrack example the input file is a pytor file.
   The times takes to load the entire file.
- For cnvpytorTrackVCF example, the input file a vcf file (Its slow)
   This one takes times as it processes an entire VCF file and show read depth and BAF information. Fo

jrobinso:
OK, I will update. 

It takes 3 minutes to load the first track,  longer to load the second.   Is this what you expect? 

Also do you really want to open the example at that specific locus?   The page is basically blank, no instructions to the end user or explanation.   This will be the first impression many people get of this track.

jrobinso:
By blank I mean there is no explanatory text,  not the blank track due to the scaling problem.

jrobinso:
I'm hand merging due to merge conflicts.  If further pushes are needed please rebase your branch from mine first.

arpanda:
There may be a download involve and that's probably increase the load time for the example pytor file. 
Size of the example pytor file is ~ 177 MB and vcf file is ~ 30 MB.

I should reduce the file size of the example pytor file and will create one, and put the file in the same path. The locus for the examples were updated to chr16 and it has few cnv event.  

Currently no further update needed. Do I need to rebase the branch ?

jrobinso:
Only if we need to update further.   My copy of the branch has the fixes for the unit tests and something there caused a merge conflict.

PR #1620, opened 2023-03-27
take ViewLimits into account when coloring featureTrack

jvolkening:
The purpose of this PR is to fix featureTrack shading when `useScore` and `viewLimits` headers are provided. `igv.js` already parses the `viewLimits` tag in BED headers, but then uses the `this.config.min` and `this.config.max` values in the actual color calculations. I wasn't able to figure out where these values are coming from, but they're not from the `viewLimits` parsing code, which sets the `dataRange` property of the track.

I wasn't sure how to modify the test suite to accommodate this change.

### DIFF
--- a/js/feature/featureTrack.js
+++ b/js/feature/featureTrack.js
@@ -454,8 +454,8 @@ class FeatureTrack extends TrackBase {
             color = IGVColor.addAlpha(color, feature.alpha)
         } else if (this.useScore && feature.score && !Number.isNaN(feature.score)) {
             // UCSC useScore option, for scores between 0-1000.  See https://genome.ucsc.edu/goldenPath/help/customTrack.html#TRACK
-            const min = this.config.min ? this.config.min : 0 //getViewLimitMin(track);
-            const max = this.config.max ? this.config.max : 1000 //getViewLimitMax(track);
+            const min = this.dataRange.min ? this.dataRange.min : 0 //getViewLimitMin(track);
+            const max = this.dataRange.max ? this.dataRange.max : 1000 //getViewLimitMax(track);
             const alpha = getAlpha(min, max, feature.score)
             feature.alpha = alpha    // Avoid computing again
             color = IGVColor.addAlpha(color, alpha)


jrobinso:
I understand the intent but I don't think this is the right solution.   "dataRange" is really used for wig type tracks, and is changed continuously if autoscale is on.    "viewLimits" is also a "wig" track parameter, not defined for bed files (annotation tracks) at UCSC.   In fact the "score" field that "useScore" refers to in the track line is defined to have a range of 0-1000.

All this is to say its complex to be compatible with the UCSC track line definitions, but we do stretch them obviously.    I think your intent is correct, setting a "min/max" range via the trackLine viewLimits should be equivalent to setting min/max via the config.   Those are static settings, dataRange is dynamic.   I will look into alternative implementations.

jvolkening:
Okay, thanks, I will close and open a "feature request" issue.

Just to note, it seems to work this way already in Java IGV, where if I set "useScore" and "viewLimits" in a BED track line, features are shaded according to the score column and using the specified viewLimits as the fully-saturated bounds. I think this is the relevant code in `IGVFeatureRenderer.java`:

```java
        if (track.isUseScore()) {
            float min = getViewLimitMin(track);
            float max = getViewLimitMax(track);

            float score = feature.getScore();
            float alpha = Float.isNaN(score) ? 1 : getAlpha(min, max, feature.getScore());
            color = ColorUtilities.getCompositeColor(color, alpha);
        }
```
and earlier when parsing the track line:
```java
else if (key.toLowerCase().equals("viewlimits")) {
    String[] limits = value.split(":");

    try {
        if (limits.length == 2) {
            float min = Float.parseFloat(limits[0].trim());
            float max = Float.parseFloat(limits[1].trim());
            trackProperties.setMinValue(min);
            trackProperties.setMaxValue(max);
        } else if (limits.length == 3) {
            float min = Float.parseFloat(limits[0].trim());
            float base = Float.parseFloat(limits[1].trim());
            float max = Float.parseFloat(limits[2].trim());
            trackProperties.setMinValue(min);
            trackProperties.setMidValue(base);
            trackProperties.setMaxValue(max);
        }
    } catch (NumberFormatException e) {
        log.error("viewLimits values must be numeric: " + value);
    }
}
```

In the current `igv.js` code:
```javascript
                case "viewlimits":
                    if (!this.config.autoscale) {   // autoscale in the config has precedence
                        tokens = properties[key].split(":")
                        let min = 0
                        let max
                        if (tokens.length == 1) {
                            max = Number(tokens[0])
                        } else if (tokens.length == 2) {
                            min = Number(tokens[0])
                            max = Number(tokens[1])
                        }
                        tracklineConfg.autoscale = false
                        tracklineConfg.dataRange = {min, max}
                    }
```

it does seem to be populating the `dataRange` property with these track line values, and also explicity turning off autoscale if they are specified, which suggested to me that somewhere along the line this might have been the intended behavior.

jrobinso:
Yes autoscale is disabled if you set viewLimits (or config.min/max),  but it can be re-enabled from the menu.  This is for wigTracks, viewLimits and dataRange do not really apply to annotation (BED) tracks unless by accident, so to speak, as in this case where we are using them for score shading.    

In the UCSC definitions, which we are playing with here,  viewLimits only apply to wig tracks, and useScore to the score field which according to their specs has a fixed range of 0-1000.   We are merging the concepts here, and using viewLimits to specify a scoreRange rather than a vertical data range axis.  That's o.k. because we already do both for config.min and config.max, but its why this is a bit messy.

jrobinso:
OK I think I see the right solution,  just after line 293 in trackBase.js add the following

```
  tracklineConfg.min = min
  trackLineConfig.max = max

```
 That should be all you need to do.   I can do this but you would have to test it anyway.

jrobinso:
This makes the trackLine setting equivalent to a config setting IFF the config setting is not used.   Config settings take precedence over trackLine settings.

jrobinso:
@jvolkening never mind, a little more complicated than that.  I'm on it now.

jrobinso:
OK I just pushed a change to allow viewLimits to modify the score range for alpha shading bed tracks.

jvolkening:
Works great...many thanks for adding this feature so quickly!

PR #1622, opened 2023-03-28
decode all URI-encoded BED attributes

jvolkening:
This PR changes the behavior of BED `gffTags` attribute parsing, removing the character subset limits of the GFF3 spec in favor of URI-decoding all encoded values in the attributes. This mirrors the behavior of standalone IGV.

One additional change was made in order to also decode the attribute key (previously, only attribute value was decoded). The GFF3 spec makes no distinction between the two in discussions of encoding and escaping characters.

Several testing modifications were made to encompass these changes.


jrobinso:
Well I sat down to do this and its done!  Thanks.

Issue #1609, opened 2023-03-17
For htsget variants source and visibilityWindow = -1 IGV get data only for one chromosome


RoadRoller:
igv.js version: 2.13.11

Steps to reproduce:
1. configure variants track with htsget source and visibilityWindow = -1

```
{
        name: 'My variants',
        url: <URL_TO_HTSGET_SERVER>,
        type: "variant",
        format: "vcf",
        sourceType: 'htsget',
        visibilityWindow: -1,
      }
```
2. Open IGV with track and select some chromosome - variants displayed as expected
3. Switch to another chromosome - results are empty, expected variants for selected chromosome

jrobinso:
Unfortunately I do not have access to an htsget server anymore,  the public endpoint from GA4GH has been broken for over a year.   So I'm not able to support htsget at this time.

PR #1645, opened 2023-05-24
Set queryable to true for HTSGET source

RoadRoller:
It's fix for https://github.com/igvteam/igv.js/issues/1609. 

For VCF from HTSGET source, only first query for some region get data from server. All subsequent requests try to get data from cahce, but cache not contains that data.

The reason of this behaviour: in textFeatureSource.js in loadFeatures method

```
        const genomicInterval = this.queryable ?
            new GenomicInterval(queryChr, intervalStart, intervalEnd) :
            undefined
```

i.e. when later we put data into cache `genomicInterval` is `undefined`. In the cache, an `undefined` range value is interpreted as included all features.

### DIFF
--- a/js/feature/textFeatureSource.js
+++ b/js/feature/textFeatureSource.js
@@ -71,6 +71,7 @@ class TextFeatureSource {
             this.queryable = true
         } else if ("htsget" === config.sourceType) {
             this.reader = new HtsgetVariantReader(config, genome)
+            this.queryable = true
         } else if (config.sourceType === 'ucscservice') {
             this.reader = new UCSCServiceReader(config.source)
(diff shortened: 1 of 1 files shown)


jrobinso:
Yes, looks correct, thanks for the PR.   I have no way to test htsget anymore as the public reference server is no longer supported,  however I will accept this PR as it looks correct and can't harm anything else.

RoadRoller:
@jrobinso, hello! Thak you for great product IGV and for accepting PR. About the htsget server, I can advice this implementation https://github.com/ga4gh/htsget-refserver that can be launched with docker locally for testing purposes.

PR #1656, opened 2023-06-27
add option for sort by start location in the right click options

wilsonGen:
There is actually a sort by start location function, but the option is missing in the right click option menu, this changes is to add back the "start location" option in the option menu


jrobinso:
Thanks for the PR

PR #1659, opened 2023-07-03
can sort as pairs when view as pairs, sort as single reads otherwise

wilsonGen:
Edited code so that :
When view as pairs, sorting will find the pairs that overlap the position,
when not view as pairs, sorting will find the single reads that overlap the position.

This behaviour is actually same as the IGV desktop version.


jrobinso:
looks good, thanks for the PR

PR #1668, opened 2023-08-08
git issue 1665 png output

turner:
Add functionality to save igvjs browser display as PNG file.


jrobinso:
Build is failing (see above), I'm sure we can work around this.  This should be a "devDepencency", btw,  igv.js has no runtime dependencies.

turner:
Yah, there's a test that is failing, otherwise it builds ok.

Issue #1730, opened 2023-11-08
Track scroll not working in splitted regions view


RoadRoller:
Issue reproduced on demo link, steps to reproduce:
1. Open IGV by link https://igv.org/web/release/2.15.5/examples/cram-vcf.html
2. Set alignments track height to 300
3. Set location to two regions, i.e. "1:10,000-19,922 1:14,771,706-14,781,628"
4. Not all alignments fit into track windown, but there is no way to scroll down, see screenshot below:

![image](https://github.com/igvteam/igv.js/assets/740624/20a24448-5dac-4ca6-81ad-b70af6b9b8be)

RoadRoller:
@jrobinso, @turner, Hello!

I looked at this with a debugger locally and think that I found the problem: https://github.com/igvteam/igv.js/pull/1731
Please take a look.

turner:
@jrobinso I think @RoadRoller fix is correct.

turner:
closed with commit b714ae04

PR #1731, opened 2023-11-08
Fixed calculation of max viewport content height

RoadRoller:
Fixed issue https://github.com/igvteam/igv.js/issues/1730


turner:
@jrobinso I agree with @RoadRoller's fix. Take a look.

jrobinso:
Yes that's correct, this was definitely a bug. @RoadRoller thanks for the fix.

PR #1742, opened 2023-12-12
fix ROI shift issue when saving svg

wilsonGen:
Fix ROI shift issue when saving svg:

Issue:
When the IGV browser is not stick to the left of the web page, the saved SVG will have the ROI shifted:
how the IGV place in the webpage:
Screenshot 2023-12-12 122912
the output SVG from IGV:
Screenshot 2023-12-12 122941

If the IGV browser is stick to the left of the webpage like this👇, then the ROI in SVG will be normal
Screenshot 2023-12-12 163438

Work around:
Minus the columnContainer's clientX when rendering SVG rect of ROI instead of just using the regionElement's clientX

### DIFF
--- a/js/browser.js
+++ b/js/browser.js
@@ -412,7 +412,9 @@ class Browser {
             trackView.renderSVGContext(context, delta)
         }
 
-        this.roiManager.renderSVGContext(context, delta, x)
+        // ROI -> SVG
+        delta.deltaX = x
+        this.roiManager.renderSVGContext(context, delta)
 
         // reset height to trim away unneeded svg canvas real estate. Yes, a bit of a hack.
         context.setHeight(height)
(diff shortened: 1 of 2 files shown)


turner:
@wilsonGen functionally it looks correct. It might be a bit cleaner to do this:
```
In browser.js
delta.deltaX = x
this.roiManager.renderSVGContext(context, delta)

```
and in ROIManager.js the calls to fillRect look like this:
```
renderSVGContext(context, {deltaX, deltaY}) {


    context.fillRect(x-deltaX, y+deltaY, width, height)
    
    
    context.fillRect(xx-deltaX, yy+deltaY, ww, hh)
}

```

wilsonGen:
@turner Oh right, this is much cleaner, thank you for the suggestions.

jrobinso:
@wilsonGen Thanks for the fix

Issue #1749, opened 2024-01-11
A bug when using CSI index for BAM alignments


beaulian:
Hi, I think there is a bug when searching alignments using CSI index, here `overlappingBins[0]` is a tuple and in the top level, causing `lowestOffset` to be undefined, maybe it should be `ba.loffset[overlappingBins[this.depth][0]]`?

https://github.com/igvteam/igv.js/blob/77d88fac63564111462b010f739ce691ac44c321/js/bam/csiIndex.js#L155

jrobinso:
Yes something is wrong here.  I think this was ported from C code, makes my head spin a bit but I will get to the bottom of it.  Thanks for the report.

jrobinso:
OK I no longer understand `loffset`, if I ever did.   According to the spec here, https://samtools.github.io/hts-specs/CSIv1.pdf, `loffset` is 

"(Virtual) file offset of the first overlapping record".   

"record" is not defined anywhere,  but I assume it is an alignment or feature.     So if loffset is the virtual position of the first alignment/feature,  why would there be any chunks in this bin with vpointers < loff?  

The bug you have found is effectively turning off the `loffset` optimization.  I don't want to guess at how it should be fixed until I understand it.  At the moment it is a harmless bug,  efficiency aside.

jrobinso:
OK this incorrect "optimization" is removed.

beaulian:
Hi, @jrobinso 

Thanks for your reply. The bug is significant when using CSI index. Without the `loffset` to reduce the search space, CSI index needs to access large bins, causing a big performance loss.

I find a similar code in htsjdk
https://github.com/samtools/htsjdk/blob/b16416c078e2df3e3a21486207d0c0c31b293cb9/src/main/java/htsjdk/samtools/CSIIndex.java#L172

they say
> Compute 'minimumOffset' by searching the lowest level bin containing 'startPos'.

So I think `lowestOffset` should be `ba.loffset[overlappingBins[this.depth][0]]`. I would greatly appreciate it if you can fix this bug. Thanks.

jrobinso:
I would consider a PR for this if it included a unit test,  since you seem to understand how loffset should be used.   I personally do not understand it, and can't find any clear description of what it actually is.

beaulian:
Hi, @jrobinso 

Thanks for your reply. I can't find a clear description either. I try my best to give my own explanation and then submit a PR.

[...]

PR #1756, opened 2024-01-26
fix loffset optimization for CSI index

beaulian:
see #1749  for the detail discussion, the implementation refers to the implementation of htsjdk.


jrobinso:
Thanks I'll try to get to this today.

beaulian:
Thanks! @jrobinso. The test data comes from https://github.com/igvteam/igv.js/blob/82048a10b58f2387d09d27f956982a7f236c450a/examples/loadTracks.html#L80

jrobinso:
This is merged, thanks for the PR

PR #1760, opened 2024-02-01
Minor fixes ROI and WIG track

milandomazet:
Fixed two minor bugs related to ROI and WIG tracks:

### ROI
When adding custom ROI and trying to name it or switch to it the coordinates were not calculated properly for contigs with - (hyphen) in their name. 
  **Example**: If we have _contig-example01-123-156_ the split will take _contig_ as the contig name and try to convert _example01_ to int for the starting locus.

Fixed by taking the last two elements of the split results as start and end and joining the rest with - and taking the result of that as a contig name.

### WIG (and Merged track)
There is a baseline drawn when there are negative values in WIG file(s). This baseline was not taking flip y-axis into consideration i.e. it was not being moved to a new position of 0 on Y axis.

Fixed by subtracting the ratio of height from 1 in case of a flipped Y axis.

---

Sorry for putting ROI and WIG related fixes in a single pull request but they seemed like minor fixes so I did not want to overcomplicate it.

### DIFF
--- a/js/feature/wigTrack.js
+++ b/js/feature/wigTrack.js
@@ -257,7 +257,8 @@ class WigTrack extends TrackBase {
 
                 // If the track includes negative values draw a baseline
                 if (this.dataRange.min < 0) {
-                    const basepx = (this.dataRange.max / (this.dataRange.max - this.dataRange.min)) * options.pixelHeight
+                    const ratio = this.dataRange.max / (this.dataRange.max - this.dataRange.min)
+                    const basepx = this.flipAxis ? (1 - ratio) * options.pixelHeight : ratio * options.pixelHeight
                     IGVGraphics.strokeLine(ctx, 0, basepx, options.pixelWidth, basepx, {strokeStyle: this.baselineColor})
                 }
             }
--- a/js/roi/ROIManager.js
+++ b/js/roi/ROIManager.js
@@ -349,9 +349,10 @@ function createSelector(regionKey) {
 }
 
 function parseRegionKey(regionKey) {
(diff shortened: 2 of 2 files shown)


jrobinso:
Thanks for the PR, I like the parsing change for "-".    BTW I would guess contig names with hyphens are going to cause problems with a other bioinformatics tools, I personally have never seen one.  It is legal of course.

milandomazet:
No worries! I was thinking the same thing and I was wondering is it just
our custom reference that is broken. I concluded based on the SAM
specification and HLA contigs in GRCh38 (e.g. HLA-A*01:01:01:01) that it
could be a valid use case for someone else as well.

jrobinso:
Yes I agree its legal so should be supported, just unusual.  I wonder how samtools would parse the HLA contigs, not only does it have "-" but also ":".

milandomazet:
Due to a reference I was working with I got to know what samtools considers good characters:

# The regexp permitted is:
# [0-9A-Za-z!#$%&+./:;?@^_|~-][0-9A-Za-z!#$%&*+./:;=?@^_|~-]*

Copied from https://github.com/samtools/samtools/blob/develop/misc/fasta-sanitize.pl

jrobinso:
Yes that's the spec, I understand those names are legal,  but I am curious happens if you want to use "samtools view <locus>"?   What do you do for the locus string?  Normally its somtheing like "chr1:100-200".     For the HLA contig you reference this would be "HLA-A*01:01:01:01:100-200".   However if you want to see the whole contig it would be "HLA-A*01:01:01:01".   Samtools might be clever about this,  IGV would not do the right thing.

PR #1763, opened 2024-02-06
Merged track fixes/improvements

milandomazet:
PR fixing issues in Merged track.

## Popup data on click
A number of things related to Merged track popup data have been fixed.

### Popup data showing all loaded points and not being focused on what the user clicked on
**Issue**: The popup data was showing all data points which were loaded and not just those the user clicked on i.e. the click on popup behaviour was not the same as it is in the individual WIG tracks. 

Screenshot 2024-02-06 at 16 56 09

As you can see in the screenshot above the list is so long that it cannot fit into the popup when the track is zoomed out. In the screenshot below you can see how it looked like when zoomed in to the individual level. The position I clicked on is 1500 and instead of showing just that value it shows a lot of values around it.
Screenshot 2024-02-06 at 16 57 39

**Fix**: Fixed by applying the clickedFeatures function logic on Merged track and adding the track name to the features list to make sure we pair the features with the correct track name. New solution screenshots in the same order (and same click position) as above:
Screenshot 2024-02-06 at 16 55 18
Screenshot 2024-02-06 at 16 57 59

### Popup being displayed when there is no data to display at all
**Issue**:
Screenshot 2024-02-06 at 17 08 32

**Fix**: Do not display popup when there is no data to show. The popup will be displayed if there is at least one track that has data in that position.
Screenshot 2024-02-06 at 17 08 48

## Y axis values when autoscaling is on and no data is displayed
**Issue**:
When there is no data to display the maximum value was being set to -Number.MAX_VALUE.
Screenshot 2024-02-06 at 16 58 33
**Fix**:
If max value stays -Number.MAX_VALUE we set it to 100 to have 0 to 100 scale as we do in WIG tracks.
Screenshot 2024-02-06 at 16 58 16

##

### DIFF
--- a/js/feature/mergedTrack.js
+++ b/js/feature/mergedTrack.js
@@ -132,12 +132,20 @@ class MergedTrack extends TrackBase {
 
     /**
      * Returns a MergedFeatureCollection containing an array of features for the specified range, 1 for each track.
+     * In addition it contains track names in the same order.
      */
     async getFeatures(chr, bpStart, bpEnd, bpPerPixel) {
-
+        
         const promises = this.tracks.map((t) => t.getFeatures(chr, bpStart, bpEnd, bpPerPixel))
         const featureArrays = await Promise.all(promises)
-        return new MergedFeatureCollection(featureArrays)
+        
+        if (featureArrays.every((arr) => arr.length === 0)){
+            return new MergedFeatureCollection([], [])
+        }
+        else {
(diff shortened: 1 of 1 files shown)


jrobinso:
Thanks for the PR and all the work on the merged track!   Looks good.  Some tests are failing but they don't look related to this PR so I'm going to merge it and fix the tests in the merged code.

PR #1765, opened 2024-02-09
Render Amino Acid Sequences

turner:


milandomazet:
Hello @turner @jrobinso for some reason that I cannot figure out (this is a large PR and my knowledge not so much) this PR slows down loading of bigwig files by a large extent. If I checkout the commit that is prior to this PR everything loads really fast but when this one is checked out the loading of bigwig tracks takes a lot of time (in my case bigwig file size is around 250MB).

jrobinso:
Hmm, that's interesting.   I'm not seeing it, can you give us more specific steps to reproduce?   Are you zoomed all the way out, all the way in,  something in between?

milandomazet:
Ok I did more troubleshooting and it looks really strange, it seems there is something that is getting the CPU to 100% and hanging the loading. So to give you a bit of background:

- I am running this on MAC OS Monterey on an Intel Macbook Pro Mid 2015 (i7)
- I am running the server using `http-server -c-1`
- I have a reference with around 20k contigs (they are transcripts) and I am using one with around 3.5kb to do my tests
- As you can see from the screenshot below I have a BAM file track and a bigwig track. BigWig has coverage data.

The network log using commit c2f8b0925b703f3383f99d97d4a4fb600dba4179 (this PR)
Screenshot 2024-02-19 at 21 48 51

The network log using commit 2948ad3d6157463fbac203282cd8ba5e02e8a9fa (commit prior to this PR)
Screenshot 2024-02-19 at 21 49 44

As you can see there is a large gap in this PR network log. I tried to figure out what are the "opening" steps of this gap and the closing steps:

Screenshot 2024-02-19 at 21 51 52
Screenshot 2024-02-19 at 21 52 01

In addition I am attaching CPU usage with this PR:
Screenshot 2024-02-19 at 21 52 33

And with the previous commit:
Screenshot 2024-02-19 at 21 53 27

jrobinso:
Thanks for the details.   There is definitely something wrong, probably specific to your usage pattern (a transcript reference for example).   Could you possibly share this test case?  If you email igv-team@broadinstitute.org I can send you a secure dropbox link to upload it.

jrobinso:
I don't understand what you are referring to when you say 

`As you can see there is a large gap in this PR network log. `

milandomazet:
What I mean by 

> As you can see there is a large gap in this PR network log. 

is that there is a huge gap in the network time diagram. The red part in this screenshot.
Screenshot 2024-02-19 at 21 52 01

I will prepare test data and reach out to you to share it ASAP.

jrobinso:
Thanks for the data, it reproduces the problem nicely.

jrobinso:
@milandomazet Thanks again for the data, I think this is fixed with the latest commit https://github.com/igvteam/igv.js/commit/91c2ca75429d089a1ff71d0edb272eae1344168e

milandomazet:
It did fix it! Thanks a lot @jrobinso

PR #1769, opened 2024-02-21
New feature: Add copy sequence on ruler track

milandomazet:
Hello, I was in need of a feature which allows you to copy parts of the sequence without zooming into it (drops in coverage for example or something similar). This PR adds the option by basically replicating the user ROI function and instead copying the selected sequence. It can be used by holding the Option (Mac) or Alt (Windows) key and dragging over the ruler. There is a limit of 10kb but if you think we can increase this it can be done easily. Please feel free to reject this one if you think it is not a useful feature overall.


jrobinso:
Its useful for sure, but I'm unsure about the mechanism.  Specifically overloading shift-drag with a 3rd option (in addition to zoom and roi)?   Perhaps we should add this to the sequence track.

milandomazet:
I can give it a try, you mean like having a shift + scroll on a sequence track as well (as there is already a scroll with just a click and drag)? That might also be troublesome due to the proximity of ruler and sequence, but I can try it and we can see how it looks like.

helgathorv:
Another thing we might consider - in IGV desktop, the regions of interest have a feature where you can copy the sequence for the region. We could add that to the igv.js ROIs

jrobinso:
@helgathorv We probably should add that to the ROI menu, and also blat,  but its not mutually exclusive with this.    Some sort of "copy" from the sequence track makes sense.   You can already see sequence at 10kb,  so there is a visual cue there.

jrobinso:
I'm closing this, you can already copy the sequence from the sequence track (from the right click menu).

PR #1776, opened 2024-03-07
New Feature: Alignment Grouping

milandomazet:
As far as I've seen this is something that has been asked for and I was in need of it as well so I gave it a run. Please let me know what you think and check below for the things I would need help with.

### What has been done
What is implemented in this PR is (what we can call) a first version of alignment grouping that does not cover all of the things but it should cover most of the needs and with your help we should be able to make it even better (I had trouble with some of the parts due to my limited knowledge about the library). 

New option has been added to the alignment track which can be provided as following:
```
name: "Name",
type: "alignment",
format: "bam",
url: "URL",
indexURL: "INDEXURL",
group: {
    option: 'STRAND'
}
```

Current options are: STRAND, FIRST_IN_PAIR_STRAND, START, INSERT_SIZE, MATE_CHR, MQ, ALIGNED_READ_LENGTH, TAG

Menu entries have been added to the alignment track popup covering all of these as well as NONE which clears the grouping. 

A dashed line separating the groups has been added as well.

### What is left to be done 

Ordered by priority (in my opinion)
- Tests (I am not really good with JS tests as Frontend is not my main focus, but I can give it a try, would appreciate help if possible)
- Showing group names in the corner of the alignment track
- Group as pairs or making the grouping aware of paired alignments when the View as pairs option is used
- Group by Base (if we think this is needed)

When it comes to Showing group names in the corner of the alignment track I would really appreciate some help, at least a commit showing me how to add any text on the alignment track (or some component over it) which will not be moved from the corner when I move the view using click drag as this is causing me issues when I just use fill text on the alignment track.

There is a TODO for all of these things in the respective files.

### Minor fixes
There were a couple other minor fixes (some of them typos) that were fixed during this feature implementation. In addition, selecting to show soft clipped bases now preserves sorting and or grouping that was already selected (it was completely resetting the view previously).

### DIFF
--- a/js/bam/alignmentContainer.js
+++ b/js/bam/alignmentContainer.js
@@ -24,9 +24,10 @@
  * THE SOFTWARE.
  */
 import PairedAlignment from "./pairedAlignment.js"
-import {canBePaired, packAlignmentRows, pairAlignments, unpairAlignments} from "./alignmentUtils.js"
+import {canBePaired, packAlignmentRows, pairAlignments, unpairAlignments, sortAlignmentRows} from "./alignmentUtils.js"
 import {IGVMath} from "../../node_modules/igv-utils/src/index.js"
 import BaseModificationCounts from "./mods/baseModificationCounts.js"
+import BamAlignmentRow from "./bamAlignmentRow.js"
 
 
 class AlignmentContainer {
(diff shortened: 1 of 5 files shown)


jrobinso:
@milandomazet Thanks for this but it will be some time before I can get to it.

jrobinso:
@milandomazet Thanks for getting a start on this.    This was merged with other work and pulled as #1794

PR #1801, opened 2024-04-25
Change prettyPrint for small numbers - see issue #1800

### LINKED ISSUE #1800: Track axis limit label unusable when values are very low (<0.01)
I have some bigWig tracks where the maximum value is very low (on the order of 1e-7 to 1e-4). Currently, the axis limit labels are useless since these low values appear as "0.00":

Current

This works fine for most tracks that don't have data maxing-out at these low orders of magnitude. But once the axis limit is below 0.01 it starts to produce erroneous labels (for example, an actual limit of 0.005 would be displayed as 0.01, and everything below that is just 0.00).

I think a decent improvement would be keep the current implementation for values between 0.1 and 1, since two decimal digits is useful information on that range, but then add a new pretty print format for values below 0.1, e.g. `number.toExponential(1)` as shown here:

Proposed

knowah:
Change axis limit labelling so that small values (< 0.1) are displayed with exponential notation (e.g. "2.3e-3") instead of "0.00" as previously implemented.

This would fix #1800

### DIFF
--- a/js/util/paintAxis.js
+++ b/js/util/paintAxis.js
@@ -49,8 +49,10 @@ function paintAxis(ctx, width, height, colorOrUndefined) {
             return number.toFixed()
         } else if (Math.abs(number) >= 1) {
             return number.toFixed(1)
-        } else {
+        } else if (Math.abs(number) >= 0.1) {
             return number.toFixed(2)
+        } else {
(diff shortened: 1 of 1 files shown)


jrobinso:
Thanks for the PR

PR #1821, opened 2024-05-23
Guess Fasta `indexURL` from functions as well

fiskus:
Hi! Thank you again for an awesome project!

I found one bug. It's not exactly a bug, since it only occurs if you disobey the documentation. Documentation doesn't mention this, but I found that - thanks to `igvxhr` - I can provide async functions for any URL in the config. However, if I set `fastaURL` as a function and omit `indexURL` and `indexFile`, IGV.js generates a `(function() {}).fai` gibberish URL.

I understand this isn't a typical bug because the documentation doesn't mention using functions for `fastaURL`. And, I know you wanted to reconsider the whole idea of guessing that URL. However, I wanted to fix this for consistency with track URLs and to add tests, so if you decide to change this behavior, it's test covered.

Feel free to request changes to this PR.

### DIFF
--- a/js/genome/indexedFasta.js
+++ b/js/genome/indexedFasta.js
@@ -33,6 +33,25 @@ const splitLines = StringUtils.splitLines
 
 const reservedProperties = new Set(['fastaURL', 'indexURL', 'compressedIndexURL', 'cytobandURL', 'indexed'])
 
+//Tries to guess indexFile based on IGV reference.
+//Note that URLs could be async thunks.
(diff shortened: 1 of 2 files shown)


jrobinso:
Yes, this looks safe as it only comes into play if indexURL is not specified.   I would like to make indexURL required, if present, but am hesitating as igv.js has been out for a long time and has many installations.   Hopefully no one is counting on this guessing behavior, but the very need for this PR argues that some are. 

Thanks again for the PR, and especially for including tests.

Issue #1674, opened 2023-08-16
About [Color By Allele Frequency] 


zlxd76:
Hi, I need to use igv.js to create the same graphics as the IGV APP.
How should I configure it, such as [Color By Allele Frequency] in IGV APP.

zlxd76:
![igv-new](https://github.com/igvteam/igv.js/assets/104063031/33a1cc3d-12b0-4935-81dc-d357cbf53e40)

jrobinso:
That option is not supported in igv.js.

zlxd76:
OK.thanks.
So do you plan to support this option?

jrobinso:
Not at this time.   There are a number of higher priorities.   You can leave this open as a request.

PR #1834, opened 2024-06-26
variant track coloring.  Support color-by allele frequence (default),…

jrobinso:
… type and svtype, arbitrary info field


jrobinso:
Add color-by-allele frequency, type, svtype, and arbitrary info field for the variant track.  Fixes #1674

PR #1862, opened 2024-08-05
bug: SVG rendering of ideogram is not happening

turner:


js/ideogramTrack.js
```diff
@@ -129,7 +129,13 @@ async function drawIdeogram({ctx, chr, referenceFrame, genome, width, height, st
 
     IGVGraphics.fillRect(ctx, 0, 0, width, height, {fillStyle: IGVColor.greyScale(255)})
 
-    const cytobands = await genome.getCytobands(chr)
+    let cytobands
+    if (0 === genome.cytobandSource.cytobands.size) {
```
jrobinso: Let's not do tests with multiple dots like this,  there's too much knowledge of genome details.  Not all genomes will even have a cytobandSource, but regardless this kind of detailed knowledge of another objects internals shouldn't be exposed.

PR #1869, opened 2024-08-08
Fix Merged track popup bug

milandomazet:
Hello! First of all, thanks for all the work you've been putting into igv.js and all the new releases!!

One thing I noticed is a small bug in Merged Track popup which prevents it from displaying (see the screenshots).

This was probably partly introduced by myself not highlighting why I added a new parameter (trackNames) to the MergedFeatureCollection class when I previously worked on this. You've probably deleted it while cleaning the code but it is important for the popup data as it is used to display track names in the order of the actual tracks in the popup. I've reintroduced it and added a comment that should hopefully make it clearer.

Let me know if this breaks anything I am missing to see.

Screenshot 2024-08-08 at 14 18 34
Screenshot 2024-08-08 at 14 18 13

Edit: another side thing, we should perhaps think about updating eslint to newer version and then updating the ecmaVersion as it complains about static class properties as it is now.

### DIFF
--- a/js/feature/mergedTrack.js
+++ b/js/feature/mergedTrack.js
@@ -227,7 +227,14 @@ class MergedTrack extends TrackBase {
 
         const promises = this.tracks.map((t) => t.getFeatures(chr, bpStart, bpEnd, bpPerPixel))
         const featureArrays = await Promise.all(promises)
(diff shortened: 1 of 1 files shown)


jrobinso:
Thanks for the PR, that was inadvertent.

Issue #1870, opened 2024-08-08
Htsget Bam support broken since v2.9.0


jmgenomics:
Following on from my colleague's [issue](https://github.com/igvteam/igv.js/issues/1691) a while back, I have been investigating why newer versions of IGV.js are no longer compatible with our Htsget server at Genomics England.  The last working version was v2.8.6.

My investigations have uncovered that the breakage occurred when code around Htsget support was refactored in commit [bcd4720](https://github.com/igvteam/igv.js/commit/bcd4720e880ef9f5a436bc16032f11a6a8e06c6c).

I have been able to determine the exact cause of the problem and have managed to fix it locally.

In `js/htsget/htsgetBamReader.js::readAlignments()` there are 2 problems:

1.  (minor issue) - `this.header.chrAliasTable` is undefined at runtime - I'm not sure of the purpose of the code and removed it.
2.  (major issue) - the call to `BamUtils.decodeBamHeader()` is happening too early, before the data has been loaded.  This has the effect that the `header.size` value passed into `BamUtils.decodeBamRecords()` is wrong.  Moving the header decoding call to after the data has been loaded and decompressed (`BGZip.unbgzf()`), fixed the issue.  

I have attached the fixed code and a visual diff for your perusal.  I'm happy to be added as a maintainer and push my branch too.

![IGV-screenshot](https://github.com/user-attachments/assets/bfd47a72-4201-4d34-aab1-654c58e4d3a5)
[htsgetBamReader.txt](https://github.com/user-attachments/files/16550007/htsgetBamReader.txt)

jrobinso:
Unfortunately we can no longer support htsget as there are no reference test servers available, and the project itself does not seem to be active.   For example this ticket has been open for more almost 3 years, in fact there are no project updates in three years: https://github.com/ga4gh/htsget-refserver/issues/33.   

That said, your change should not affect any other code negatively so I would accept is as a PR.   I'm in general hesitant to accept PRs that I cannot personally test, but will make an exception in this special case.

jmgenomics:
Thanks @jrobinso, happy to raise a PR for this.  I'm ready to push my branch - can you add me as a collaborator to the repo please?

jrobinso:
@jmgenomics  You don't need to be a collaborator to submit a PR.

jrobinso:
[...]

PR #1872, opened 2024-08-09
Fix htsget missing chrAliasTable and decoding header after data fetch

jmgenomics:
Potential fix for issue [1870](https://github.com/igvteam/igv.js/issues/1870) reported.

- add statements back into `bamUtils.js` that were present prior to the breakage.
- re-evaluates the header after the full fetch of data, to ensure the offset is correct for decoding

### DIFF
--- a/js/htsget/htsgetBamReader.js
+++ b/js/htsget/htsgetBamReader.js
@@ -44,8 +44,8 @@ class HtsgetBamReader extends HtsgetReader {
             const ba = BGZip.unbgzf(compressedData.buffer)
             this.header = BamUtils.decodeBamHeader(ba, this.genome)
             this.chrAliasTable = new Map()
-            for (let key of Object.keys(this.header.chrAliasTable)) {
-                this.chrAliasTable.set(key, this.header.chrAliasTable[key])
+            for (let name of this.header.chrNames) {
+                this.chrAliasTable.set(name, this.genome.getChromosomeName(name))
             }
         }
 
@@ -55,6 +55,7 @@ class HtsgetBamReader extends HtsgetReader {
 
         // BAM decoding
         const ba = BGZip.unbgzf(compressedData.buffer)
+        this.header = BamUtils.decodeBamHeader(ba, this.genome)
 
         const chrIdx = this.header.chrToIndex[chr]
         const alignmentContainer = new AlignmentContainer(chr, start, end, this.config)


js/htsget/htsgetBamReader.js
```diff
@@ -55,6 +55,7 @@ class HtsgetBamReader extends HtsgetReader {
 
         // BAM decoding
         const ba = BGZip.unbgzf(compressedData.buffer)
+        this.header = BamUtils.decodeBamHeader(ba, this.genome)
```
jmgenomics, line 58: `header` needed to be decoded again due to difference in `header.size`.  If relying on the first assignment (line 45), the offset used for `BamUtils.decodeBamRecords()`, will be wrong.

js/bam/bamUtils.js
```diff
@@ -73,15 +74,21 @@ const BamUtils = {
             chrToIndex[name] = i
             chrNames[i] = name
 
+            if (genome) {
```
jmgenomics: Adding statements back in from the last working version v2.8.6 
https://github.com/igvteam/igv.js/blob/v2.8.6/js/bam/bamUtils.js#L114

js/htsget/htsgetBamReader.js
jrobinso: The change is harmless, but I would like to understand it.    Why would header.size differ? Shouldn't the header be exactly the same?

jmgenomics: The header isn't exactly the same.  We are using `samtools` in our htsget implementation and it outputs the command used in the PG header fields, which differs between fetching header only vs full data.

Here's an example diff:
header-diff

jrobinso: OK, that's reasonable.  Odd that no one has  encountered this before but there's no real evidence htsget is being used in igv.js until now.

jrobinso:
With the header.size problem fixed, I do not understand why the implementation as is does not work.   Could you elaborate on where you think the code below is failing?  `this.chrAliasTable` does not need to be referenced or defined prior to this of course, it is defined here.   

```
        if (!this.header) {
            const compressedData = await this.readHeaderData()
            const ba = BGZip.unbgzf(compressedData.buffer)
            this.header = BamUtils.decodeBamHeader(ba, this.genome)
            this.chrAliasTable = new Map()
            for (let key of Object.keys(this.header.chrAliasTable)) {
                this.chrAliasTable.set(key, this.header.chrAliasTable[key])
            }
        }
```

jrobinso:
Could you try this, in `readAlignments`?

```
        if (!this.header) {
            const compressedData = await this.readHeaderData()
            const ba = BGZip.unbgzf(compressedData.buffer)
            this.header = BamUtils.decodeBamHeader(ba, this.genome)
            this.chrAliasTable = new Map()
            for (let name of this.header.chrNames) {
                this.chrAliasTable.set(name, this.genome.getChromosomeName(name))
            }
        }
```

jrobinso:
Also revert the changes to bamUtils.

jmgenomics:
> Could you try this, in `readAlignments`?
> 
> ```
>         if (!this.header) {
>             const compressedData = await this.readHeaderData()
>             const ba = BGZip.unbgzf(compressedData.buffer)
>             this.header = BamUtils.decodeBamHeader(ba, this.genome)
>             this.chrAliasTable = new Map()
>             for (let name of this.header.chrNames) {
>                 this.chrAliasTable.set(name, this.genome.getChromosomeName(name))
>             }
>         }
> ```

Yes, this works! 

I reverted all changes, used this code and also the additional `this.header = BamUtils.decodeBamHeader(ba, this.genome)`.

jrobinso:
OK, if you can update the PR I will merge it.  Just pushing your branch should update it.

jrobinso:
OK, merged, thanks for the PR.

Issue #1881, opened 2024-09-03
contrib: devcontainer config 


PeterKnealeCMRI:
I'd like to add a devcontainer config file to make it easier to run this project locally and contribute. I've included an example of one below. After cloning the repository and opening it in VSCode the user will be asked if they with to spin up a docker container and develop within that. This means the tooling and environment used to write the application can be made consistant and reliable. You can also spin up codespaces within github and develop using an 'on demand' instance when circumstances require it.  

> What are Development Containers?
A development container (or dev container for short) allows you to use a container as a full-featured development environment. It can be used to run an application, to separate tools, libraries, or runtimes needed for working with a codebase, and to aid in continuous integration and testing. Dev containers can be run locally or remotely, in a private or public cloud, in a variety of [supporting tools and editors](https://containers.dev/supporting).

Further info: https://code.visualstudio.com/docs/devcontainers/containers

```js
// For format details, see https://aka.ms/devcontainer.json. For config options, see the
// README at: https://github.com/devcontainers/templates/tree/main/src/javascript-node
{
	"name": "Node.js",
	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
	"image": "mcr.microsoft.com/devcontainers/javascript-node:1-20-bookworm",
	// Features to add to the dev container. More info: https://containers.dev/features.
	// "features": {},
	// Use 'forwardPorts' to make a list of ports inside the container available locally.
	// "forwardPorts": [],
	// Use 'postCreateCommand' to run commands after the container is created.
	"postCreateCommand": "npm install && npm install --global http-server",
	"portsAttributes": {
		"8080": {
			"label": "http-server"
		}
	}
	// Configure tool-specific properties.
	// "customizations": {},
	// Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
	// "remoteUser": "root"
}
```

PeterKnealeCMRI:
Further to this, when setting this project up locally I noticed that the [github action](https://github.com/igvteam/igv.js/blob/master/.github/workflows/ci_build.yml#L15) is using `node-version: '19.x'`

I'm not hugely familiar with the node / js ecosystem but I think its generally considered more reliable to stick to the even numbered releases. If using the above dev container it also means you can pin the same version in both your github action and devcontainer ensuring releases are built the same locally and within the CI process.
https://nodejs.org/en/about/previous-releases#release-schedule

jrobinso:
[...]

PR #1884, opened 2024-09-04
Add dev container

PeterKnealeCMRI:
This PR adds a dev container per discussion.
Opening the repository in VS Code when this file is present will create a dokcer image, install the necessary dependencies then all development in the pre-configured environment.
https://github.com/igvteam/igv.js/issues/1881

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,22 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/javascript-node
+{
+	"name": "Node.js",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/javascript-node:1-20-bookworm",
+	// Features to add to the dev container. More info: https://containers.dev/features.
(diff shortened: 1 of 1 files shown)


jrobinso:
It looks simple enough, but there is a maintenance issue here as no one on this team use VS.

PeterKnealeCMRI:
VS Code is just one editor supporting this new(ish) standard. 
IntelliJ IDEA and others support it to if you happen to use that. The list  is here: https://containers.dev/supporting

Github itself supports it, once this file is present in the repo, you can just click on code-spaces and provision a web based version of vscode and have the same environment installed.. I've found for a few bits of bioinformatics pieces of work with large  fasta files etc, the cloud based editors are actually much quicker to work with because of their proximity to resources like s3 etc.

image

I dont think there is a large maintenance burden, it just makes pulling down a repo and working on it that bit simpler without worrying about what versions of node/npm/packages are installed.

jrobinso:
It is a burden for us.  If we commit to having this in our repository we commit to testing and maintaining it.     I don't understand the comment on package versions, those are all nailed down in package.json.   We should add a note on required node version to the readme, to be honest I don't know what that is.   Probably anything later than 12.

PeterKnealeCMRI:
Thats okay, I appreciate you have limited time. 
Thanks for your efforts on the library - its very useful.

PR #1899, opened 2024-10-06
BigWig track: Add none window Function

nivm:
Hi All, 
We would like to add to igv.js a none window function. 

This functionality is available for wig source and in IGV desktop. IGV Desktop screenshot:

Screenshot 2024-10-06 at 10 11 33 PM

Example from the example files: 

Screenshot 2024-10-06 at 10 16 38 PM


jrobinso:
OK.   I'm sure you are aware this will freeze your browser if you naively try it on a large bigwig.   We should probably introduce a visibility window to allow this, but we don't on desktop.   So user beware.

PR #1905, opened 2024-10-24
Navbar refactor

jrobinso:


jrobinso:
Refactor navbar code - introduce class for navbar.

PR #1910, opened 2024-11-05
fix none windowFunction for wig track in bedgraph format

1pratham:
Hello All,

https://github.com/igvteam/igv.js/pull/1899 added none window function for wig track. However, it doesn't work with bedgraph format with error about `Unkown window function: None` when selecting "all" chromosome

image

This PR fixes it

image


nivm (approved):

jrobinso:
Thanks

PR #1927, opened 2025-01-15
Add API, prototype demo UI for interactive filtering

eweitz:
This updates the IGV API to enable point-and-click filtering for genomic features.  That makes analysis richer and nimbler.

### Impact
Previously, if users wanted to e.g. subset a VCF track to show only variants that are SNPs with AF > 0.05, then they would need to generate a new track.  That requires significant effort writing code, managing resulting data, and configuring IGV.

Now, it's feasible to develop an interactive filtering UI that makes such detailed genome analysis quick and easy in IGV.  Below is a filtering UI that shows how a user can load an arbitrary VCF and query genomic features in view by any INFO field.  

Categorical fields like "Variant type" or "Ancestral allele" are filterable via checkboxes.  Numeric fields like "Allele frequency" or "Read depth" are filterable via range inputs and sliders.  The _distribution_ of field values can also be assessed at a glance.  Checkboxes show counts beside them, and range inputs show a histogram above them.  The demo UI shows utility, but it's not robust.  It simply show that the API is adequate for the core filtering UI.

The client presents faceted search controls using only pre-existing data; no file-specific configuration or processing is needed to enable filtering.  In addition to VCFs as shown below, these API updates also enable developing filtering UIs for BED data.

Here's how it looks!

https://github.com/user-attachments/assets/1dda5538-3a19-4a57-bc14-edc139ed9032

The demo UI code is a first pass at porting a mature React IGV filtering UI to plain JavaScript.  That aids interoperability and reusability.  The React IGV filtering UI is from [Single Cell Portal](https://singlecell.broadinstitute.org), particularly SCP's [multiome filtering](https://singlecell.broadinstitute.org/single_cell/features/2024-11-21-visualize-and-filter-multiome-atac-data) functionality.  A [related announcement](https://singlecell.broadinstitute.org/single_cell/features/2024-01-04-filter-plotted-cells-for-a-deeper-dive-into-study-data) shows a richer demo of the faceting widgets in the video above, including possible enhancements like sparkbars.


jrobinso:
Thanks @eweitz, this looks straightforward.  The demo looks great.   I'll get to this asap, no later than next week.

eweitz:
Thanks Jim!  I just pushed some UI refinements and polish that improve on the [earlier](https://github.com/user-attachments/assets/7f440f05-c05e-4558-8cbc-e02bd3e7d7ae) filtering UI .

js/index.js
jrobinso, line 1: Changes to index.js need to be rolled back

js/trackViewport.js
jrobinso, line 1: We shouldn't need this event any longer,  filters are not triggered on load (nor should they be).

js/index.js
eweitz: Done in 6240157.  I'd implemented this locally, but overlooked pushing.

PR #1962, opened 2025-04-12
fix race condition

guoci:
During the incremental update of `GenomeUtils.KNOWN_GENOMES`, the async function `igvxhr.loadJson` is being executed. This leads to a race condition if we create multiple IGV browsers.

### DIFF
--- a/js/genome/genomeUtils.js
+++ b/js/genome/genomeUtils.js
@@ -11,7 +11,7 @@ const GenomeUtils = {
 
         if (!GenomeUtils.KNOWN_GENOMES) {
 
-            GenomeUtils.KNOWN_GENOMES = {}
+            let table = {}
 
             const processJson = (jsonArray, table) => {
                 jsonArray.forEach(function (json) {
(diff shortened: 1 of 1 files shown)


jrobinso:
Thanks for the PR!  However I don't see how this solves the issue,  and in fact it looks like it could make it worse.   Setting the global GenomeUtils.KNOWN_GENOMES immediately in any "IGV browser" should prevent the block from being executed from the others. 

Could you elaborate, what exactly is the race condition?  I am missing something.

guoci:
We need to keep `GenomeUtils.KNOWN_GENOMES` undefined until we have a fully formed list for it, then set it at one point.

If not, another async function sees that `GenomeUtils.KNOWN_GENOMES` is defined but may see only the empty/partial list.

You can also see the difference after you made the change at version 3.2.5. Version 3.2.4 works fine.
https://igv.org/web/release/3.2.4/examples/interact.html
https://igv.org/web/release/3.2.5/examples/interact.html

jrobinso:
OK, got it, but with the change as coded all "browser" objects could be loading GenomeUtils simultaneously.   This is harmless, I guess, if a bit wasteful of bandwidth.

jrobinso:
As this was a breaking bug for anyone with multiple browsers I have merged and am releasing it as 3.2.6.   The genome json is loaded multiple times,  but the file is small and that is harmless compared to 3.2.5 behavior.   I might revisit this later to prevent the multiple loads.    Thanks again for the PR.

guoci:
[async-lock](https://www.npmjs.com/package/async-lock) is a solution to that, too bad Javascript doesn't have it as a built in.

PR #1977, opened 2025-05-14
Support for bigPsl translated alignment data from UCSC.

rmhubley:
Displays aligned amino-acids at zoomed level and flags characters that do not correspond to the translated sequence.  Added a "bigPsl.html" example that loads the UCSC Uniprot track data.

### DIFF
--- a/examples/bigPsl.html
+++ b/examples/bigPsl.html
@@ -1,0 +1,98 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+    <meta content="IE=edge" http-equiv="X-UA-Compatible">
+    <meta content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no" name="viewport">
+    <meta content="" name="description">
+    <meta content="" name="author">
+    <link href=../img/favicon.ico rel="shortcut icon">
+    <title>hg38 test</title>
+
+</head>
+
+<body>
+
+<h1>bigPsl Uniprot Track from UCSC -- with amino acid level alignment details</h1>
+
+<p>
+    <b>Look for console errors and compare with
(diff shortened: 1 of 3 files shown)


rmhubley:
Of course I submitted that too quickly.  I found a problem with how I am interpreting the bigPsl data (at least for this particular track),  I have a request into UCSC for clarification.  I will close this out for now and reopen once I get clarity.

jrobinso:
I have this on my todo list also, along with other "big" formats.  PSL (plain text format) is already handled and rendered correctly, AFAIK,  you shouldn't have to make any changes to "renderFeature".

jrobinso:
The general approach to take here is to add a decoder to "bigwig/bbDecoders.js" for bigPsl, along the lines of the bigInteract decoder already there.   The decodePsl function in feature/decode/ucsc.js could be a model.  In an idea world that function would be used directly.   Once the bb file is decoded directly everything else, including rendering, should work.

rmhubley:
I started off by writing a decodeBigPsl function but then I noticed that you also have a mechanism to auto-detect format extensions for bigBed files using autoSQL.  This was sufficient for me to extend your gene feature renderer to support the special case of a translated alignment and the display of the amino-acid details at high zoom levels:

Screenshot 2025-05-14 at 4 34 10 PM

I had to extend the renderer for this as I wanted it to display the amino acid sequences from the PSL file rather than the translated sequence. Happy to refactor if you have a larger vision for how this should be done.  I will also be working on porting over our UCSC RepeatMasker Track so that I have access to it through IGV.  It would be helpful to have a better understanding of how you would like to see extensions coded.  

I fixed the problem I identified earlier.  You can try it out here: http://www.repeatmasker.org/~rhubley/igv/igv.js/examples/bigPsl.html

jrobinso:
@rmhubley Looks great,  open a PR when you have it worked out,  I don't know when I'll get to it but maybe next week.

PR #1987, opened 2025-05-21
Add genomicLocation to the "trackclick" event

milandomazet:
Hello! A small PR aimed at the trackclick event and data available in the handler: **genomicLocation** has been added to the event handler for the **trackclick** event.

The reasoning behind this was mainly aimed at the bigwig tracks which can have a location range in the popup even though we click on a single location (a bigwig file can have ranges in it for example 100-200 23). I thought it might be useful having the exact location where the user clicked as this might be used in the handler itself. The new field was added as optional so we do not raise typing issues with current implementations users of igv.js might have. We could add the whole clickState as well but it seems to me that it would be an overkill.

Another fix I did while working on the above is related to the EvenReturn type for trackclick event. For this one I guess it was only boolean at some point and was not changed once string was allowed in the popup function.

As always, feel free to reject this one if you don't find this change useful.

### DIFF
--- a/js/igv.d.ts
+++ b/js/igv.d.ts
@@ -528,6 +528,7 @@ export namespace BrowserEvents {
                 T extends "trackclick" ? (
                         track: Tracks.Track,
                         popoverData?: Record<string, string>,
+                        genomicLocation?: number
                     ) => EventReturn<T> :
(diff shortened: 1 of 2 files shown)


jrobinso:
Looks fine, but keep in mind that unless zoomed in a single pixel can span many bases,  so what you get for genomicLocation does not in general imply a particular base was "clicked" on.

PR #1993, opened 2025-05-26
Update genomes.json BACKUP_URL

dnil:
We are having some intermittent issues with the `genomes.json` file from `igv.js` at the moment in Stockholm. A directly configured `genomeList` works fine with `loadDefaultGenomes: false`, but while investigating, it was noticed that the references repo directory structure has been rearranged a bit, so the `igv.js` `BACKUP_GENOMES_URL` is off. It looks like the current live one is `https://raw.githubusercontent.com/igvteam/igv-data/main/genomes/web/genomes.json`.
Or should we be using something like https://s3.us-east-1.amazonaws.com/igv.org.genomes/genomes.json instead?

### DIFF
--- a/js/genome/genomeUtils.js
+++ b/js/genome/genomeUtils.js
@@ -3,7 +3,7 @@ import {convertToHubURL} from "../ucsc/ucscUtils.js"
 import {loadHub} from "../ucsc/hub/hubParser.js"
 
 const DEFAULT_GENOMES_URL = "https://igv.org/genomes/genomes.json"
-const BACKUP_GENOMES_URL = "https://raw.githubusercontent.com/igvteam/igv-genomes/refs/heads/main/dist/genomes.json"
+const BACKUP_GENOMES_URL = "https://raw.githubusercontent.com/igvteam/igv-data/main/genomes/web/genomes.json"
 
 const GenomeUtils = {
 


jrobinso:
I'm not sure which URL you are referring to.  There are some intermittent  issues at the moment with https://igv.org, a ticket has been created with the Broad Institute.   However you should not get a 404 for that URL.

I am also not sure why you would get a 404 for `https://raw.githubusercontent.com/igvteam/igv-data/refs/heads/main/genomes/web/genomes.json`.   There are few resources more stable than github.

As you seem to be only using one or a few genomes the most stable long term option is to host the assemblies you use on your own server.

cariaso:
>I'm not sure which URL you are referring to. 

I was speaking of
>the official url is still https://igv.org/genome/genomes.json . 

it is this which is now consistently giving me  404 after a long wait.

```
$ curl https://igv.org/genome/genomes.json
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>404 Not Found</title>
</head><body>
<h1>Not Found</h1>
<p>The requested URL was not found on this server.</p>
</body></html>

$ curl -I https://igv.org/genome/genomes.json
HTTP/1.1 404 Not Found
Date: Mon, 26 May 2025 19:37:32 GMT
Server: Apache
Content-Type: text/html; charset=iso-8859-1
```




is the choice of
https://raw.githubusercontent.com/igvteam/igv-data/refs/heads/main/genomes/web/genomes.json
vs
https://igv.org/genome/genomes.json
configurable from the options param? seems not. 


is there some config based on raw.githubusercontent.com/igvteam that avoids the currently problematic
https://igv.org/genome/genomes.json

PR #2054, opened 2025-11-25
Fix condition for multiWig container check

kpalin:
Replaced missing "equals" function with "==".

Apparently, at least for the track hub I attempted to use, property "container" is a string and does not have ".equals" method.  Replacing it with '==' "works for me".


jrobinso:
Thanks for the PR!

PR #2077, opened 2026-03-16
adds lineWidth property; adds example

ynedelec:
Hello,

For some visualisation, it would be nice to change the thickness of the line when `graphType` is set to `line`.

I've made some modifications to the code that renders the wig track and it looks correct:
image

The line width can be set using the `lineWidth` property in the track definition. An example is available in the `examples` folder.

Does that sound like a valuable addition ?

Cheers!


jrobinso:
Yes, good addition.

Issue #2084, opened 2026-06-12
displayName sanity check


gsfk:
I've been using igv.js with TypeScript (thanks for the type declarations!) but have lately been tripped up by a mysterious required parameter.

A parameter called `displayName` appears in an example of a custom search service ([code](https://github.com/igvteam/igv.js/blob/518f27e63e2b43d229b27e349c7947c7c9ab10c4/dev/misc/customSearchService.html#L30-L33) | [live web page](https://igv.org/web/release/3.5.0/examples/customSearchService.html)), and in the [TypeScript declarations file](https://github.com/igvteam/igv.js/blob/518f27e63e2b43d229b27e349c7947c7c9ab10c4/js/igv.d.ts#L509-L513) as a required field, but this parameter doesn't seem to exist in the code (it's not mentioned in the docs either).

- The docs for search config are [here](https://igv.org/doc/igvjs/#Browser-Creation/#search-object-details), and they don't mention `displayName`. 

- For the code, you can see the search config params consumed [here](https://github.com/igvteam/igv.js/blob/518f27e63e2b43d229b27e349c7947c7c9ab10c4/js/browser.js#L194-L205), although this includes two parameters missing from the docs (`geneField` and `snpField`). It does not include  `displayName`. 

There is no exact match for "displayName" anywhere in the code; if you ignore case you can find `chrDisplayName`, `getChromosomeDisplayName()` and `#getGoogleDriveDisplayName()`, which appear to be unrelated. I looked quickly to see if this was simply an old parameter that got deprecated, but couldn't find anything. 

Despite this, a custom search service written in TypeScript requires you to use `displayName`, because it's typed as a required field, while most fields actually used for a search service are missing:

https://github.com/igvteam/igv.js/blob/518f27e63e2b43d229b27e349c7947c7c9ab10c4/js/igv.d.ts#L509-L513

It looks like the typing was based on the example linked above, since it includes the same three parameters. An easy workaround is to provide dummy values for displayName in your TypeScript code, since the value is ignored. 

I offer a fix in #2085.

jrobinso:
Fixed

PR #2085, opened 2026-06-12
Fix TypeScript type declaration for custom web search config

gsfk:
Corrects the TypeScript type declarations for a custom search service. 

Current typing looks like this:

    search?: {
        url: string;
        chromosomeField: string;
        displayName: string;
    };

This omits multiple search config fields such as `startField`, `endField`, `resultsField`, etc (docs [here](https://igv.org/doc/igvjs/#Browser-Creation/#search-object-details), although docs are incomplete. You can see the config params consumed [here](https://github.com/igvteam/igv.js/blob/518f27e63e2b43d229b27e349c7947c7c9ab10c4/js/browser.js#L194-L205), or see [this test](https://github.com/igvteam/igv.js/blob/518f27e63e2b43d229b27e349c7947c7c9ab10c4/test/testSearch.js#L18-L27)). The typing seems to be based on [this example](https://github.com/igvteam/igv.js/blob/518f27e63e2b43d229b27e349c7947c7c9ab10c4/dev/misc/customSearchService.html), but the example doesn't seem to be correct. See issue [#2084](https://github.com/igvteam/igv.js/issues/2084) for a longer discussion. 

The biggest obstacle for development is that the current typing has both `chromosomeField` and `displayName` as required fields, when:
- chromosomeField is optional
- displayName isn't used anywhere (discussed in issue [#2084](https://github.com/igvteam/igv.js/issues/2084))

This pr adds in the expected fields, making them optional where it's clear in the code that they are optional. ~This includes the parameter `timeout` which is consumed in a [different part of the code](https://github.com/igvteam/igv.js/blob/518f27e63e2b43d229b27e349c7947c7c9ab10c4/js/searchFeatures.js#L68) than the rest of the config.~

### DIFF
--- a/js/igv.d.ts
+++ b/js/igv.d.ts
@@ -515,7 +515,6 @@ interface CreateOptExtras {
         resultsField?:  string;
         geneField?: string;
         snpField?:  string;
-        timeout?: number;
     };
 }
 
--- a/js/igv.d.ts
+++ b/js/igv.d.ts
@@ -508,8 +508,14 @@ interface CreateOptExtras {
 
     search?: {
         url: string;
-        chromosomeField: string;
-        displayName: string;
+        coords?: 0 | 1;
+        chromosomeField?: string;
+        startField?: string;
+        endField?:  string;
+        resultsField?:  string;
+        geneField?: string;
+        snpField?:  string;
+        timeout?: number;
     };
 }
 


gsfk:
After a few checks, I removed the timeout parameter. This value exists only in the [default search config](https://github.com/igvteam/igv.js/blob/518f27e63e2b43d229b27e349c7947c7c9ab10c4/js/searchFeatures.js#L9-L14) that is used when you don't use a custom search service. It's currently impossible to set for a custom search service, since the value is not read when the search config is ingested. This seems like an oversight with an easy fix, but I'll leave it for a separate pr.

jrobinso:
Thanks for the PR

PR #2097, opened 2026-09-18
Fix TLEN read coloring if mate is slow to load

dnil:
Hi,

first, thank you so much for making igv.js!

We have been having rare issues where color alignment by `TLEN` (or `unexpectedPair`) fails. Opening one side of a deletion can show nicely coloured reads, but when visiting the other side, the color disappeared. For the vast majority of events with improper read pairs, the color is as expected. But when it happens, it is striking and affects many pairs. Toggling show mate in Split View could help - or trigger the issue. 

It has, for good and bad, been hard to reproduce. When moving the same cram file from the production and staging servers to local storage, the issue disappeared, even though the very same range request service was used to deliver them.  I have been unable to make a small cram to reproduce the issue. I hope you have some ideas here, and that you have more insight into how this might happen, and how to test it properly!

I'm guessing this is the most likely problem area:
https://github.com/igvteam/igv.js/blob/0fe2e26c0ba3414057ddcb021e527db28db57690/js/bam/alignmentTrack.js#L1454

When adding debug statements before the `fragmentLength` check, the issue resolves, and the TLEN color pattern is restored. It also resolves if we skip the `alignment.mate` check before `fragmentLength` as in this PR.

There are some images over at https://github.com/Clinical-Genomics/scout/issues/4548 and https://github.com/Clinical-Genomics/scout/pull/6568 that might possibly shed some light.


jrobinso:
This is curious.  The alignment.mate and alignment.fragmentLength fields are populated synchronously at decode time from the alignment record itself, whether the mate loads or not.  So mate load time or failure would not affect this field.  This is a hard one so I turned "Claude" up to xhigh on model Opus 5 to review it.  It thrashes about and contradicts itself, but there are some valid points below.   I have fixed and pushed the problem noted wrt js/cram/fileHandler.js.   This might fix your issue, but I doubt it, it was a bug.

Is this bug occuring in "multi locus view"?  Your observation that adding a log statement fixes it points to a race or timing condition with loading.  The tlen min and max parameters are computed based on the distribution of values in the first 100 reads loaded.   So its possible the log statement is delaying reading on one side vs the other and we are getting different tlen values.   This can be tested by explicitly setting max tLen in the configuration for the track.  In fact this is recommended regardless as it takes the stochastic element out wrt tLen.  Set this in the track config with property "maxTLEN".   You can set "minTLEN" also if you like.  Report back here if this appears to fix the bug.

In the meantime I will look into a more robust method to compute maxTLEN and minTLEN.   It's going to always have some variability, we can't read the entire CRAM to compute these values.


-------------- Claude output follows ---------------------


I don't want to take this patch as-is though, for two reasons: I think it regresses some other reads, and I don't think it actually fixes the case you're hitting.

### Why the `alignment.mate` check can't just be dropped

`alignment.mate` and `isMateMapped()` look redundant but answer different questions. `isMateMapped()` is `(flags & 0x8) === 0` — a flag test that's only meaningful when `READ_PAIRED` (0x1) is also set. `alignment.mate` is a data-presence test. Removing the second one makes two populations newly reachable, and both arrive with `TLEN == 0`:

* **Unpaired reads.** No 0x1 and no 0x8, so `isMateMapped()` returns `true`; no mate object (`bamUtils.js` only builds one when `mateChrIdx >= 0`); TLEN 0. They now hit `0 < minTemplateLength` and get painted `smallTLENColor`. Any file mixing pairs with orphan reads is affected. A pure single-end file escapes only by accident, because `percentile([], 0.1)` returns `undefined` and both thresholds end up falsy — so this won't show up in the obvious test.
* **Interchromosomal reads with no mate object** — i.e. the case this PR is about. The chr-color branch is now gated on `mate`, but the `else if` fall-through isn't, so those reads fall through to the thresholds. Per the SAM spec TLEN is 0 when the segments map to different references, so they also come out `smallTLENColor`.

`Number.isFinite` is what lets both through: `Number.isFinite(0)` is `true`, but 0 is precisely SAM's sentinel for "template length unavailable". So instead of leaving the read uncolored, we assert "small insert". For your use case I think that's a worse failure than the one you reported — an uncolored discordant read gets noticed, a blue one gets believed.

### Why I don't think it fixes the underlying problem

In `cramReader.js` we do `record.templateLength || record.templateSize`, and in cram-js those two are populated on mutually exclusive paths that *each also populate the mate*: a detached record reads MF/NS/NP (mate) and TS (`templateSize`); an intra-slice mated record gets both `mate` and `templateLength` from `associateIntraSliceMate`. So when `record.mate` is missing, both template fields are `undefined` too, `Number.isFinite(undefined)` is `false`, and the patch is a no-op on exactly the reads it was written for.

Also, `alignment.mate` is assigned synchronously when the record is decoded — nothing fills it in later, and the color function only runs on records a resolved `loadFeatures()` has already cached. So there's no window where the mate is "slow to load," which makes me think the debug-statement observation and the color coming back are both downstream of something else.

### If you'd like to keep the intent

Using TLEN when the mate object is absent is reasonable — it just needs truthiness rather than finiteness, plus an `isPaired()` test:

```js
case "tlen":
case "fragmentLength": {
    const mate = alignment.mate
    if (alignment.isPaired() && alignment.isMateMapped()) {
        if (mate && mate.chr !== alignment.chr) {
            color = getChrColor(mate.chr)
        } else if (mate || alignment.fragmentLength) {   // 0/undefined TLEN + no mate => no information
            const tlen = Math.abs(alignment.fragmentLength)
            if (this.minTemplateLength && tlen < this.minTemplateLength) {
                color = this.smallTLENColor
            } else if (this.maxTemplateLength && tlen > this.maxTemplateLength) {
                color = this.largeTLENColor
            }
        }
    }
    break
}
```

That keeps everything the current code colors, adds the mate-less-but-real-TLEN case you're after, and leaves TLEN 0 alone. (Worth wrapping the case in braces either way — `const` in a bare `case` leaks into the whole `switch` scope.)

### Where the real bug probably is

  Their strongest clue is the one that points away from getAlignmentColor entirely: same file, same range service, bug vanishes when served from local storage, and a debug statement makes it go away. That's a
  truncated-or-stale-bytes shape, not a color-logic shape.

  js/cram/fileHandler.js is worth a look on its own merits. Cache.get fetches [s, e) and then registers new Chunk(s, e, buffer) without checking that buffer.byteLength === e - s. If a proxy or edge cache returns a
  short body, contains() will report a hit for bytes that were never received, and buffer.slice() clamps silently rather than throwing — so read() quietly returns fewer bytes than requested. That path is already
  exercised routinely at EOF, where the over-read past the end always comes back short.

  A cheap diagnostic for the contributor: log record.mate, record.templateLength, record.templateSize, record.cramFlags (detached = 0x2, mate-downstream = 0x4) and record.mateRecordNumber in decodeCramRecord for the
  affected reads. mate undefined with both template fields undefined means cram-js never associated the intra-slice mate and this PR can't help; mate undefined with templateSize present means the contradictory
  detached case above.

jrobinso:
@dnil I think https://github.com/igvteam/igv.js/pull/2100 might actually fix your issue.  Setting a maxTLEN property might as well, both fixes are assuming that the computation of maxTLEN was the root cause.

dnil:
Thank you @jrobinso! Yes, much more credibly than the present PR. Good analysis. I discounted the limit calculation since the first example had the some color disappear when opening the mate in a split view, but maybe they are recomputed? Or we jumped directly to the Split View. Anyway, the current master looks like a distinct improvement with direct implications for the issue, that also resolves the current example. Looking forward to the next release!

jrobinso:
@dnil The TLEN fix has been released as 3.8.9

PR #2102, opened 2026-09-22
Fix stale fillstyle applied to unmodified bases

levend1s:
Noticed this bug while using igv-reports and saw strange modification patterns when using baseModificationThreshold. This should fix a stale variable being applied to unmodified bases.


jrobinso:
Thanks for the PR

